PR #15, opened 2026-05-21
Make BCQuality an additive knowledge layer with agent findings

JesperSchulz:
## Why

Today the DO output contract effectively requires every finding to cite a BCQuality knowledge article: leaf sub-skills only emit findings they can back with a `references[]` entry, and the orchestrator strips findings without one. That makes BCQuality a *replacement* for the agent's review judgement instead of an additive layer on top of it. Things the agent spots on its own but that no knowledge file covers yet get silently dropped.

The companion change in `microsoft/BCAppsBCQuality` already teaches the orchestrator to render unbacked findings as long as they carry a marker. This PR makes BCQuality formally produce them.

## Approach

- **Pick one canonical marker:** `from-sub-skill: "agent"`. It reuses an existing DO field, so no schema additions, and the orchestrator already accepts it.
- **Scope agent findings to super-skills.** Leaf sub-skills stay strictly knowledge-driven (one knowledge subset per leaf, citation required). The `al-code-review` super-skill gains a self-review pass that runs alongside its sub-skill rollup.
- **Validate before surfacing.** The self-review pass checks every agent-identified candidate against the BCQuality knowledge already loaded for the task. If a knowledge file matches the candidate is upgraded to a knowledge-backed finding and merged/deduplicated; if a knowledge file explicitly contradicts the candidate it is suppressed; otherwise it surfaces as an agent finding.
- **DO contract changes (strictly additive):**
  - `references: []` is permitted only when `from-sub-skill: "agent"`.
  - Agent-finding `id` is a slug prefixed `agent:` (mirrors the existing `<from-sub-skill>:` rule).
  - `confidence` for agent findings is capped at `medium` (no citation, no authoritative basis for `high`).
  - `message` must be self-contained because there is no knowledge-file footer to fall back on.

## Files

- `skills/do.md` - new "Agent findings" rules + updated `references`/`from-sub-skill` semantics.
- `microsoft/skills/review/al-code-review.md` - new self-review step in Action; example agent finding in the worked example.
- `agent-consumption.md` - new "Knowledge-backed and agent findings" section.
- `README.md` - one-paragraph callout pointing at the marker.

## Backward compatibility

Strictly additive. Every existing producer keeps producing valid output, and every existing consumer keeps consuming it unchanged. DO is a stable contract that requires both maintainers to approve, so calling that out explicitly: this is a permitted-shape extension, not a breaking change.

## Validation

`.github/scripts/validate_frontmatter.py` runs clean (0 errors, 0 warnings). The validator is structure-only and does not lint JSON output, so the change is documentation-and-contract level by design.

## Coordinated change

Lands together with the orchestrator change in `microsoft/BCAppsBCQuality` (`tools/Code Review/scripts/Invoke-CopilotPRReview.ps1`) which already accepts the marker, renders agent findings with a distinct label, and exposes `COPILOT_REVIEW_AGENT_MINIMUM_SEVERITY` for an independent severity floor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,8 @@ Action skills follow a four-step pattern:
 
 Every action skill produces output in a common format that orchestrators can consume without skill-specific parsing. The format is JSON and includes an `outcome` (so a clean run, a not-applicable skill, and a partial failure are all distinguishable), `findings` (what the skill observed), structured `references` back to the knowledge files that informed each finding, per-finding `confidence`, and a `suppressed` list recording any knowledge files overridden by layer precedence. This contract is defined in the Action Skill meta-skill so that orchestrators and action skills remain independently evolvable.
 
+BCQuality is an **additive** knowledge layer: it augments the agent's review judgement, it does not replace it. Super-skills (such as `al-code-review`) run a self-review pass alongside their sub-skills and surface concerns the agent identified on its own, marked with `from-sub-skill: "agent"` and an empty `references: []` so consumers can render them distinctly from knowledge-backed findings. See [agent-consumption.md](agent-consumption.md) and [`skills/do.md`](skills/do.md) for the full contract.
+
 The meta-skills in `/skills/` define this pattern. Every concrete action skill follows it.
 
 For the end-to-end flow — from orchestrator trigger through to how output reaches developers — see [agent-consumption.md](agent-consumption.md).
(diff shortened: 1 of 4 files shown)


JeremyVyska:
This seems really good based on initial read-through.  I like the additive adjustment approach here, as as models increase in quality and detection, it will adapt the results.

I will be curious if we can even create some scenarios, not unlike BC Benchmark and CentralGauge, where we can eval over time what knowledge becomes part of the models FROM this project, so we can slowly "prune" leaves.

Let's merge it up and I'll try some test runs with this ASAP.

PR #23, opened 2026-05-28
Define mandatory suggested-code classes

JesperSchulz:
Follow-up to microsoft/BCQuality#22.

The synthetic benchmark after #22 showed the model followed the new contract but remained too conservative:

```text
Suggested-code diagnostics: mechanical-looking=13 with-suggestion=2 omission-reason=11 missing-both=0
```

So #22 fixed silent omission, but not actionability. This PR removes ambiguity by defining explicit **must-suggest classes** where omission is not allowed if the target `location` points to the affected line/range and the replacement can be expressed as one contiguous block.

## Must-suggest classes

- Missing `OptionCaption` on an `Option` field — add `OptionCaption` matching the existing members.
- Missing `ToolTip` on a page field/control — add a concise `ToolTip` property.
- Uppercase reserved keywords or simple formatter-only whitespace/casing defects — emit the formatted line/range.
- `Count() > 0` or repeated `Count()` used for existence — replace with `not IsEmpty()` or cache the count when it is also returned.
- Unreachable/dead code after an unconditional terminator — remove the unreachable contiguous block.
- Over-broad AL permission tokens where the required token is obvious (`rimd` on read-only access → `r`).
- Missing `DataClassification` where the correct value is explicit in the article or obvious from the field.
- Local `Label` moves when the target range includes both the local declaration and object-level insertion point; otherwise use omission reason.

## Changes

- `skills/do.md` documents the must-suggest table.
- All six AL leaf skills reference the must-suggest classes.
- `al-code-review.md` references the same classes for rolled-up and self-review findings.

## Companion PR

microsoft/BCAppsBCQuality#39 updates CI diagnostics to report must-suggest violations separately from generic mechanical-looking omissions.


JesperSchulz:
Closing per design feedback: the generic DO contract should stay domain-agnostic and must not encode AL-specific examples like OptionCaption/ToolTip. The broader suggested-code/omission-reason contract from #22 remains the right abstraction layer.

PR #30, opened 2026-06-15
Add community knowledge file: integration event parameter is a breaking change

jeffreybulanadi:
Adds `integration-event-parameter-is-a-breaking-change` to `community/knowledge/events/`.

Adding a parameter directly to an existing `[IntegrationEvent]` breaks every subscriber immediately — no deprecation period, no warning, no grace. The correct pattern is to add a new overloaded event alongside the original, mark the original `[Obsolete]`, and raise both during the transition period.

Includes `.good.al` and `.bad.al` sibling samples.

### DIFF
--- a/community/knowledge/events/integration-event-parameter-is-a-breaking-change.md
+++ b/community/knowledge/events/integration-event-parameter-is-a-breaking-change.md
@@ -1,0 +1,26 @@
+---
+bc-version: [all]
+domain: events
+keywords: [integration-event, publisher, subscriber, breaking-change, parameter, obsolete]
+technologies: [al]
+countries: [w1]
+application-area: [all]
+---
+
+# Adding a parameter to an existing integration event is a breaking change
+
+## Description
+
+Every codeunit that subscribes to an `[IntegrationEvent]` must match the publisher's exact parameter signature. Adding a parameter to an existing event immediately breaks every subscriber — they fail to compile the moment the parameter is added to the publisher. This affects all consumers of the event, including those in other extensions the author does not control.
+
+## Best Practice
+
+Add a new event with the extended signature alongside the original. Keep the original event and mark it `[Obsolete(...)]` so existing subscribers continue to compile and authors have time to migrate. The new event name should reflect the addition (for example, append `WithShipmentNo` or increment a suffix). Raise both events during the transition period.
+
+See sample: `integration-event-parameter-is-a-breaking-change.good.al`.
+
+## Anti Pattern
+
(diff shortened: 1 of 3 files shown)


community/knowledge/events/integration-event-parameter-is-a-breaking-change.md
```diff
@@ -0,0 +1,26 @@
+---
+bc-version: [all]
+domain: events
+keywords: [integration-event, publisher, subscriber, breaking-change, parameter, obsolete]
+technologies: [al]
+countries: [w1]
+application-area: [all]
+---
+
+# Adding a parameter to an existing integration event is a breaking change
```
JesperSchulz, line 10: It is not a breaking change? You can add parameters all you like, without breaking existing subscribers. You cannot change existing parameters, neither can you insert new ones in the middle of the parameter list. But you can append. I don't think this article is right.

JesperSchulz (changes requested):

JesperSchulz (changes requested):
The best-practice recommendation is reasonable, but the core premise of this file, that adding a parameter to an existing integration event breaks every subscriber at compile time, isn't correct, and the `.bad.al` sample demonstrates a compile error that doesn't actually happen.

In AL, a subscriber can declare a **left-anchored subset** of the publisher's parameters and omit trailing ones. So when `.bad.al` appends `ShipmentNo` to the end of the publisher:

```al
procedure OnAfterPostSalesOrder(var SalesHeader: Record "Sales Header"; ShipmentNo: Code[20])
```

this subscriber still compiles fine — it just ignores the new trailing parameter:

```al
local procedure HandleAfterPostSalesOrder(var SalesHeader: Record "Sales Header")
```

There's no AL0306 here. Appending a parameter at the end is actually the *safe* case, so the example contradicts itself.

The real story is more nuanced and worth capturing instead:
- Existing subscribers keep compiling but **silently never receive the new argument**, so they miss the data — that's the actual risk.
- The genuinely breaking edits are **removing** a parameter, **changing** a parameter's name/type, **reordering**, or **inserting** a parameter *before* ones a subscriber already uses — those break the left-anchored name/type match.

I'd keep the "add a new overload and mark the original `[Obsolete]`" guidance, since that's good advice, but fix the rationale and rework `.bad.al`. As written, this would teach the agent incorrect compiler behavior.

Refs: [Subscribing to events](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-subscribing-to-events) (Example 1 shows subset matching) and the [EventSubscriber attribute](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/attributes/devenv-eventsubscriber-attribute) reference.

JesperSchulz:
Covered in a different PR.

PR #31, opened 2026-06-15
Add community knowledge file: avoid raising events inside try functions

jeffreybulanadi:
Adds `avoid-raising-events-inside-try-functions` to `community/knowledge/events/`.

When an `[IntegrationEvent]` is raised inside a `TryFunction` body, any error a subscriber throws is silently caught and discarded. The subscriber's logic fails but the caller never knows. The fix is to raise the event before entering the TryFunction, so only the operation that genuinely needs error isolation lives inside it.

Includes `.good.al` and `.bad.al` sibling samples.

### DIFF
--- a/community/knowledge/events/avoid-raising-events-inside-try-functions.md
+++ b/community/knowledge/events/avoid-raising-events-inside-try-functions.md
@@ -1,0 +1,26 @@
+---
+bc-version: [all]
+domain: events
+keywords: [tryfunction, integration-event, subscriber, error-handling, silent-failure, event-publisher]
+technologies: [al]
+countries: [w1]
+application-area: [all]
+---
+
+# Do not raise integration events inside a TryFunction
+
+## Description
+
+A `TryFunction` catches all errors — including errors thrown by event subscribers. When an `[IntegrationEvent]` is raised inside a `TryFunction` body, any error a subscriber raises is silently swallowed by the TryFunction's error boundary. The subscriber's logic fails, the caller sees no error, and the calling code continues as if nothing happened. Subscribers have no way to signal failure to the caller.
+
+## Best Practice
+
+Raise the integration event before entering the TryFunction scope. The event and its subscribers execute outside the error boundary, so subscriber errors propagate normally to the caller. Move only the operation that genuinely needs error isolation (such as an HTTP call or a posting step) inside the TryFunction.
+
+See sample: `avoid-raising-events-inside-try-functions.good.al`.
+
+## Anti Pattern
+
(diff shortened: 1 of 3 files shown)


community/knowledge/events/avoid-raising-events-inside-try-functions.md
```diff
@@ -0,0 +1,26 @@
+---
+bc-version: [all]
+domain: events
```
JesperSchulz, line 3: I wonder if events are a domain, or if it should rather be extensibility?

JesperSchulz (commented):
This one's solid — the core lesson is correct and worth having. Raising an integration event synchronously inside a `[TryFunction]` does mean a subscriber's `Error()` gets caught by the try boundary and turned into a `false` return, so it never reaches the caller. Raising the event before entering the try scope is the right fix. 👍

Two things I'd tighten up before merging:

**1. The `Cancel := true` comment in `.bad.al` contradicts its own code.**

```al
Cancel := false;
OnBeforeSubmitPayment(PaymentAmount, Cancel);
if Cancel then
    exit;
```

The comment says *"A subscriber setting Cancel := true is also lost when TryFunction returns false."* But trace it: a subscriber that sets `Cancel := true` without erroring hits `if Cancel then exit`, so the try method returns **true**, not false — `Cancel` is honored, not lost. The only path that returns `false` is a subscriber raising an error, and in that case `if Cancel` is never reached, so `Cancel` is irrelevant. So that sentence doesn't describe a real failure mode here. (It also leans on "var changes are discarded on a false return," which isn't documented — the [try methods doc](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-handling-errors-using-try-methods) actually says database writes inside a try method are *not* rolled back.)

If anything, the real second footgun in that path is the opposite: a subscriber setting `Cancel := true` makes the try method `exit` and return true, so `SubmitPayment` treats a silently-skipped payment as success. I'd just drop the misleading line and keep the file focused on the one clear lesson — the swallowed `Error()`.

**2. Minor wording:** the `.md` says *"A `TryFunction` catches all errors."* I'd soften that to "catches errors thrown during its execution" — "all errors" overstates it.

Optional nit: in `.good.al`, `HttpClient.Get()` returns a Boolean rather than throwing on a transport failure, so the try method is really isolating the explicit `Error('HTTP %1'…)`, not `Get()` itself. The pattern's still fine — the "operation that can fail" framing is just a touch loose.

Net: good content, just fix the `Cancel` comment and the "all errors" wording.

community/knowledge/events/avoid-raising-events-inside-try-functions.md
JesperSchulz: We've now made it a domain.

Issue #34, opened 2026-06-23
Add the "develop" half: generation output contract, scaffolding skills, golden templates, and anti-hallucination reference


JesperSchulz:
## Context

Today every skill in BCQuality is a **reviewer**, and `skills/do.md` defines `findings-report` as the only output kind. To become the "one-stop shop" that lets agents — especially cheap/fast models — **develop** BC apps (not just review them), BCQuality needs a generative half. This is the foundational issue of the initiative; the testing/telemetry/etc. issues depend on the generation contract landing here.

## Why it matters (admission test)

Cheap models fail mostly by **hallucinating structure and signatures**. Golden templates + curated reference knowledge convert "invent the skeleton" into "fill a known-good skeleton," which is exactly where small models win or lose.

## Scope / checklist

**Output contract & skills**
- [ ] Define a generation output contract (e.g. `generation-result`) alongside `findings-report` in `skills/do.md`. ⚠️ Contract change — requires both-maintainer approval per the README.
- [ ] Add authoring/scaffolding action skills under `microsoft/skills/author/`: new table & table extension, page/page extension, API page, permission set, install codeunit, upgrade codeunit, interface, event subscriber, report.
- [ ] Teach Entry to route "generate/scaffold" goals to these skills.

**Golden templates & project conventions**
- [ ] `app.json` schema knowledge: `idRanges`, `dependencies`, platform/application/runtime versions, `features`, `target`.
- [ ] Object file naming (`<Name>.<Type>.al`), object-ID ranges, affixes/prefix, folder layout, `.alpackages`/symbols, rulesets.

**Anti-hallucination platform reference** (application-object reference lives in the functional-domains issue)
- [ ] Curated signatures + semantics of core platform methods: `FindSet`/`SetLoadFields`/`CalcFields`/`CalcSums`, `InStream`/`OutStream`/`Blob`, `JsonObject`/`JsonToken`/`JsonArray`, `HttpClient`/`HttpContent`.

**New knowledge domains needed to generate well (currently uncovered)**
- [x] Error handling — `ErrorInfo`, collectible & actionable errors, `IsHandled` pattern.
- [x] Interfaces & enum-with-implementation.
- [x] Events & subscribers — publishing integration/business events, manual vs static, `IsHandled`.
- [ ] Pages/Reports deep — current `actionref`/promoted-action syntax (old training data is wrong), ControlAddin, RDLC/Word layouts, request pages.
[...]

Issue #3, opened 2026-04-19
Layer precedence ordering disagrees between READ and Entry


JeremyVyska:
cc @JesperSchulz

Noticed while walking the routing flow for a community-layer PR: the two contracts that specify layer precedence disagree on the middle tier.

### READ (`skills/read.md`, knowledge files)

> 1. `/custom/` wins over `/microsoft/` and `/community/`.
> 2. `/microsoft/` wins over `/community/`.

Ordering: **custom > microsoft > community**.

### Entry (`skills/entry.md`, action skills)

> Skill layer precedence is `/custom/` over `/community/` over `/microsoft/` — the same ordering READ defines for knowledge files.

Ordering: **custom > community > microsoft**. The inline claim that this matches READ is incorrect.

### Why it matters

- Both orderings are individually defensible. Knowledge prefers authority (`/microsoft/` wins). Skills plausibly prefer local customization (`/community/` overrides Microsoft defaults). But the two contracts shouldn't silently disagree, and Entry's comment shouldn't assert an alignment that doesn't hold.
- This becomes a real bug when a community contributor ships a skill with the same `id` as a Microsoft skill: under Entry's current rule the community skill wins, which is a surprising outcome given READ's framing.

### Options

1. **Align on READ's ordering** (`custom > microsoft > community` for both) — skills layer behaves the same as knowledge; Microsoft-endorsed skills always win unless a consumer fork overrides in `/custom/`.
2. **Keep Entry's ordering and fix the text** — keep skills as `custom > community > microsoft` with an explicit rationale, and correct READ's claim of equivalence.
3. **Document both and explain the split** — if the asymmetry is intentional, call it out in both files.

No preference from me — just flagging before the first community skill lands, since a community `al-performance-review` under today's rule would supersede the Microsoft one.

PR #43, opened 2026-06-23
Add events knowledge domain and review leaf skill

JesperSchulz:
## Summary

Adds a new **`events`** knowledge domain covering AL **events & subscribers**, fully wired into the AL review pipeline. This is increment #3 of the knowledge-seeding initiative, following `error-handling` and `interfaces`. The change is **purely additive — no contract change** (`skills/read.md|write.md|do.md|entry.md` untouched).

### 3 atomic articles (each with `.good.al` + `.bad.al` samples)
Under `microsoft/knowledge/events/`:

1. **Use the `IsHandled` pattern to make base behaviour overridable** — publisher exposes `[IntegrationEvent(false, false)] local procedure OnBeforeX(…; var IsHandled: Boolean)`; the routine raises it and guards `if IsHandled then exit;`; a subscriber sets `IsHandled := true`. Bad: no `OnBefore`/`IsHandled` hook (not extensible), or a missing `if IsHandled then exit;` guard.
2. **Publish thin `OnBefore`/`OnAfter` integration events to expose extension points** — raise thin events at operation boundaries passing `var Rec`/`var IsHandled`, with empty publisher bodies. Bad: business logic inside the `[IntegrationEvent]` publisher body, or no extension points at all.
3. **Choose static vs manual subscribers deliberately; bind manual subscribers with `BindSubscription`** — static `[EventSubscriber]` for always-on; `EventSubscriberInstance = Manual` + `BindSubscription`/`UnbindSubscription` for scoped behaviour. Bad: static subscriber for always-on side effects, or a manual subscriber bound (and held on a `SingleInstance` global) and never unbound.

### Leaf review skill
- `microsoft/skills/review/al-events-review.md` — modeled on `al-error-handling-review.md`: 5 sections (`Source`/`Relevance`/`Worklist`/`Action`/`Output`), sources the `events` domain, worklist tokens (`IntegrationEvent`, `BusinessEvent`, `EventSubscriber`, `IsHandled`, `BindSubscription`, `UnbindSubscription`, `EventSubscriberInstance`, `OnBefore`, `OnAfter`, `Manual`), populated + empty-corpus `findings-report` examples.

### Wiring (additive)
- `microsoft/skills/review/al-code-review.md` — added the leaf to the frontmatter `sub-skills` list and the `## Source` bullet list; appended `, events` to the line-6 `description` parenthetical.
- `README.md` — leaf-skill count `seven → eight`; appended `, events` to the domain list.

## API verification
All AL event syntax was **verified against Microsoft Learn** (not training recall): `IntegrationEvent`/`BusinessEvent` attributes and empty publisher bodies; the exact `[EventSubscriber(ObjectType, ObjectId, EventName, ElementName, SkipOnMissingLicense, SkipOnMissingPermission)]` order; `EventSubscriberInstance = Manual;`; `BindSubscription`/`UnbindSubscription`; and the `var IsHandled: Boolean` + `if IsHandled then exit;` convention. Object IDs use **50220–50235** (within the assigned 50220–50239 range; no collision with error-handling 50180–50191 or interfaces 50200–50219). The `.al` samples are **demonstration-only and not compiled by CI**.

## Validation
- `python .github/scripts/validate_frontmatter.py` → **0 errors, 0 warnings**
- `pwsh .github/scripts/Test-KnowledgeIndex.ps1` → **PASSED** (163 articles, deterministic, full coverage, selection inputs intact)

## Notes
- May need a **trivial rebase if PR #42 (interfaces) merges first** — both touch the same `al-code-review.md` sub-skills/Source/description lines and the README count/domain line. Resolve by keeping both leaves and setting the README count to the correct total.

Part of #34.


microsoft/knowledge/events/add-new-event-parameters-at-the-end.md
```diff
@@ -0,0 +1,26 @@
...
+technologies: [al]
+countries: [w1]
+application-area: [all]
+---
+
+# Add new event parameters at the end
+
+## Description
+
+Adding a parameter to an existing event publisher changes its signature, and every subscriber must be updated to match. Appending the new parameter at the end of the parameter list keeps the change easy to review and minimizes churn: existing subscribers still bind to the leading parameters, and the diff is a single clean addition. Inserting a parameter in the middle shifts every following argument, makes diffs noisy, and is error-prone to reconcile across many subscribers — a subscriber that compiles can still receive the wrong values because positions moved. New parameters belong after the existing ones.
```
AleksandricMarko: this is more our way of doing than limitation, as it is possible to add parameter wherever and it will not break anyone, but it is a way easier to track changes when adding new parameters to the end

microsoft/knowledge/events/do-not-add-ishandled-to-an-existing-event.md
```diff
@@ -0,0 +1,26 @@
...
+technologies: [al]
+countries: [w1]
+application-area: [all]
+---
+
+# Do not add IsHandled to an existing event
+
+## Description
+
+Adding a `var IsHandled: Boolean` parameter to an event that already shipped without one is a breaking contract change. The signature changes, so existing subscribers no longer match and silently stop firing until they are updated, and the event's meaning shifts from "notify" to "overridable" — a semantic the original subscribers never agreed to. The safe move is to leave the existing event untouched and introduce a new `OnBefore…` event carrying `IsHandled` at the point you want to make overridable. Existing subscribers keep working against the original event; new subscribers opt into the override seam through the new one.
```
AleksandricMarko: Not breaking change but change the purpose of event so existing subscribers (on regular event) could become useless. This is introduced recently, seems like partners are not big fan of this "rule" and from time to time we let it go, but in general make sense to limit this kind of change (even it is not breaking)

microsoft/knowledge/events/add-new-event-parameters-at-the-end.md
JesperSchulz: But wouldn't you consider this being a best practice? The alternative is of course to move this knowledge out of the Microsoft layer and into the custom layer...

microsoft/knowledge/events/do-not-add-ishandled-to-an-existing-event.md
JesperSchulz: I would surely consider this a best practice that should be followed.

microsoft/knowledge/events/add-new-event-parameters-at-the-end.md
AleksandricMarko: the best practice definitely, just "and every subscriber must be updated to match" is mine concern

microsoft/knowledge/events/name-events-by-publisher-position.good.al
```diff
@@ -0,0 +1,36 @@
+// Demonstration-only AL. Not compiled by CI; illustrates the article.
+codeunit 50255 "Event Naming Good Sample"
```
AleksandricMarko, line 2: maybe it makes sense adding example for triggers (both tables and reports)

microsoft/knowledge/events/initialize-ishandled-to-false-before-publishing.md
```diff
@@ -0,0 +1,26 @@
...
+technologies: [al]
+countries: [w1]
+application-area: [all]
+---
+
+# Initialize IsHandled to false before publishing
+
+## Description
+
+A routine that raises an `OnBefore…` integration event with a `var IsHandled: Boolean` parameter passes that variable in by reference, so its incoming value decides whether the default logic is skipped. A freshly declared Boolean starts as `false`, but the same variable is frequently reused to raise several events in one routine, and after the first raise it may already be `true`. Assigning `IsHandled := false;` on the line immediately before every raise makes the control flow deterministic and self-documenting, and prevents a stale `true` from silently suppressing logic the author never meant to make skippable. Generated code often reuses one `IsHandled` across several raises without resetting it.
```
AleksandricMarko, line 14: from time to time i see OnBefore procedure events where isHandled is not declared explicitly, which is fine, as nothing should go before that event (just mentioning to avoid review comment for these kind scenarios)

microsoft/knowledge/events/do-not-bypass-critical-operations-with-ishandled.md
```diff
@@ -0,0 +1,26 @@
...
+technologies: [al]
+countries: [w1]
+application-area: [all]
+---
+
+# Do not bypass critical operations with IsHandled
+
+## Description
+
+The IsHandled override pattern lets a subscriber skip the guarded code entirely. That is acceptable around a pure, side-effect-free calculation, but dangerous around critical operations — posting, ledger-entry creation, number-series consumption, and referential-integrity or permission validation. Wrapping those in `OnBeforeX(…; var IsHandled); if IsHandled then exit;` lets any subscriber silently suppress them, risking imbalanced ledgers, orphaned documents, skipped permission checks, or duplicated numbers — corruption that surfaces far from the subscriber that caused it. Make the calculation overridable, not the commit: expose the value computation through IsHandled, or offer a regular `OnAfter…` event to adjust results, while the critical work runs unconditionally.
```
AleksandricMarko: this one is tricky as it is hard to provide good enough explanation what is critical operation we should not skip, as in general we can skip everything, but make sense at least to forbid skipping parts of code which can`t stay as independent peace

microsoft/knowledge/events/add-new-event-parameters-at-the-end.md
JesperSchulz: OK, let's get rid of that part then!

AleksandricMarko (approved):

PR #56, opened 2026-06-26
community(ui): SetSelectionFilter on list pages silently scopes to cursor row when no explicit selection

DmitryKatson:
## Summary

Adds a community knowledge article documenting a non-obvious `SetSelectionFilter` behaviour that LLMs demonstrably get wrong when generating batch list-page actions.

## What the article covers

`CurrPage.SetSelectionFilter(Rec)` leaves `MarkedOnly = false` and writes a primary key filter for the cursor row only when the user has not explicitly multi-selected records. A batch action that passes this result directly to a processing codeunit silently processes one row instead of all visible records — no error is raised.

The fix is to check `MarkedOnly` after `SetSelectionFilter` and fall back to `Copy(Rec)` when it is false, which restores the full page view including all active filter groups.

## Files added

- community/knowledge/ui/set-selection-filter-list-scope.md — knowledge article
- community/knowledge/ui/set-selection-filter-list-scope.good.al — correct pattern
- community/knowledge/ui/set-selection-filter-list-scope.bad.al — anti-pattern

## Admission test

Would a capable LLM generating a batch list-page action get this wrong without this article? Yes — LLMs routinely write CurrPage.SetSelectionFilter(Rec); Codeunit.Run(...) without knowing that the cursor-only case produces a single-row PK filter rather than the full page view.


microsoft/knowledge/ui/set-selection-filter-list-scope.md
```diff
@@ -0,0 +1,26 @@
+---
```
JesperSchulz, line 1: This looks like knowledge which is generally applicable and useful to all. It should go straight into the Microsoft layer.

JesperSchulz (changes requested):

microsoft/knowledge/ui/set-selection-filter-list-scope.md
DmitryKatson: Done ;)

JesperSchulz (approved):

JeremyVyska (approved):

PR #62, opened 2026-06-29
Add testing knowledge: UI handlers, table relations, asserterror, fixtures (P1+P2)

JesperSchulz:
## Summary

Adds six BC-platform-specific **testing**-domain knowledge articles under `microsoft/knowledge/testing/`, each as a trio (`<slug>.md` + `<slug>.good.al` + `<slug>.bad.al`) — 18 files total. Every article clears the README admission test: a modern LLM reviewing or generating BC test code would get the mechanic *wrong* without it. None drift into generic test-writing advice (AAA, naming, one-assert-per-test).

> Originally drafted in `community/knowledge/testing/`; relocated to the **microsoft/** layer per maintainer request (pure `git mv`, no content changes).

## The six articles & admission-test rationale

**P1 trio**
- **`ui-calls-require-test-handlers`** — UI calls (`Confirm`, `Message`, `Page.RunModal`, `StrMenu`, `Notification.Send`, …) need a registered handler + `[HandlerFunctions]`, or the test aborts with an *unhandled-UI runtime error* — not an assertion failure. LLMs routinely omit handlers.
- **`tablerelation-requires-prerequisite-records`** — `Validate`/`Insert(true)` enforces `TableRelation`; fixtures must be built parent-before-child or the test aborts on a relation error. LLMs assemble test data bottom-up.
- **`handlers-enqueue-never-assert`** — handlers run in a separate context; an assertion inside a handler can be swallowed, so the test passes while broken. Capture via `LibraryVariableStorage.Enqueue`, assert in the body, finish with `AssertEmpty`. Non-obvious platform behavior.

**P2 trio**
- **`handlerfunctions-attribute-must-match-ui-path`** — two-sided rule: a missing handler → unhandled UI; an unused listed handler → "handler function was not executed". Both are runtime failures. LLMs get the bidirectional contract wrong.
- **`asserterror-needs-expectederror-and-code`** — bare `asserterror` passes on *any* error; pair it with `Assert.ExpectedError` + `Assert.ExpectedErrorCode` to verify the *specific* failure. LLMs treat `asserterror` as sufficient.
- **`use-library-codeunits-for-test-fixtures`** — `LibrarySales`/`LibraryInventory`/… create valid records (number series, mandatory fields, relations); hand-rolled `Init`/`Insert` skips required setup and rots as the schema evolves. LLMs invent `Init`/`Insert` fixtures.

Articles mirror the format/frontmatter of the existing `microsoft/knowledge/testing/transactionmodel-attribute-governs-test-transactions` article (6-field frontmatter, `domain: testing`, `bc-version: [all]`, `technologies: [al]`, `countries: [w1]`, `application-area: [all]`; code in sibling `.al` files only). Each `.md` is 24–26 lines (well under the 100-line cap).

## Validator outputs

```
$ python .github/scripts/validate_frontmatter.py --root .
Validator: 0 error(s), 0 warning(s)        (exit 0)

$ pwsh .github/scripts/Test-KnowledgeIndex.ps1 -Root .
BCQuality index: 198 article(s). ...
Knowledge-index check PASSED: 198 articles, deterministic, full coverage, selection inputs intact.   (exit 0)

$ pwsh ./tools/Build-KnowledgeIndex.ps1
BCQuality index: 198 article(s). Index: .../knowledge-index.json
198        (exit 0)
```

## Article count invariant

Baseline on `main`: **192** articles → after adding 6: **198** articles. Delta = **+6**. (The community→microsoft relocation is a move, so it leaves the count unchanged at 198.)

## CODEOWNERS note

These files now live under `microsoft/knowledge/testing/`, which **CODEOWNERS gates with the testing domain experts** (@nikolakukrika @ventselartur @pchriste-microsoft-com). The PR therefore auto-requests them and is expected to sit **BLOCKED on the 2-approval ruleset** until they review. **CODEOWNERS was not edited.**

---
Do not merge — open for maintainer review.


microsoft/knowledge/testing/asserterror-needs-expectederror-and-code.bad.al
```diff
@@ -0,0 +1,18 @@
+codeunit 50409 "Test AssertError Bad"
+{
+    Subtype = Test;
+
+    [Test]
+    procedure BlankNameIsRejectedWithSpecificError()
+    var
+        Customer: Record Customer;
+    begin
+        Customer.Init();
```
nikolakukrika, line 10: It would be good to have an init pattern so we do not confuse the agent with bad sample

microsoft/knowledge/testing/asserterror-needs-expectederror-and-code.good.al
```diff
@@ -0,0 +1,23 @@
...
+        Customer: Record Customer;
+    begin
+        Customer.Init();
+        Customer.Name := '';
+
+        // [WHEN] a mandatory field is blank
+        asserterror Customer.TestField(Name);
+
+        // [THEN] verify the SPECIFIC failure — message and code — not just "any error"
+        Assert.ExpectedError('Name must have a value');
```
nikolakukrika: it should be a label

microsoft/knowledge/testing/asserterror-needs-expectederror-and-code.good.al
```diff
@@ -0,0 +1,23 @@
...
+        Assert.ExpectedErrorCode('TestField');
```
nikolakukrika: This example is bad as we are hardcoding the value inline.
There is Assert.ExpectedTestFieldError
I will move the same method to the Library Assert so it can be reused.

nikolakukrika: We should not hardcode the labels and codes. Ideally we should add more to the library so we are protected against future changes.

microsoft/knowledge/testing/asserterror-needs-expectederror-and-code.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Description
+
+`asserterror` passes when the guarded statement raises any error at all. That is too permissive for a negative test: a typo, a missing setup record, or a permission failure all raise errors, so a bare `asserterror` can go green while never exercising the rule it claims to verify — false confidence that the validation works. Constrain it. `Assert.ExpectedError(text)` checks the message of the error that was actually raised, and `Assert.ExpectedErrorCode(code)` checks its error code. Together they assert that the specific failure occurred, turning "something went wrong" into "the right thing went wrong for the right reason".
+
+## Best Practice
+
+Follow every `asserterror` with a verification of the error it expects: assert the action raises, then `Assert.ExpectedError` with the message — or a stable substring of it — and, where the code is known, `Assert.ExpectedErrorCode` (for example `'TestField'` for a mandatory-field check). Prefer matching on a code or an invariant fragment over the full localized sentence so the test survives caption changes without going blind to the wrong error.
```
nikolakukrika: Please add the text about the library and the helper methods that are there. Also the library should be extended or your own library should be introduced so we can change the behavior easy in the future.

microsoft/knowledge/testing/handlerfunctions-attribute-must-match-ui-path.good.al
```diff
@@ -0,0 +1,34 @@
...
+    local procedure RunPostingThatConfirmsAndMessages()
+    begin
+        if Confirm('Post this document?', false) then
+            Message('Posting completed.');
+    end;
+
+    [ConfirmHandler]
+    procedure ConfirmHandlerYes(Question: Text; var Reply: Boolean)
+    begin
+        Reply := true;
```
nikolakukrika: This is still a bad example. We should use the LibraryVariableStorage.Dequeue and assert if the question is as we expect it. Only part of the question should be checked, not entire text. Then we should dequeue the reply and provide it. This way we ensure that we get the right handler and that it is raised on the exact amount of times we expect it to be raised.

microsoft/knowledge/testing/handlerfunctions-attribute-must-match-ui-path.good.al
```diff
@@ -0,0 +1,34 @@
...
+    end;
+
+    [MessageHandler]
+    procedure PostMessageHandler(Message: Text)
+    begin
+        LibraryVariableStorage.Enqueue(Message);
```
nikolakukrika: This is wrong code. We should dequeue and Assert. Part of the string should be used.

nikolakukrika: We should use Assert.ExpectedMessage, Assert.ExpectedConfirm, Assert.ExpectedStrMenu

microsoft/knowledge/testing/handlerfunctions-attribute-must-match-ui-path.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Best Practice
+
+List precisely the handlers the scenario triggers, comma-separated, in any order. When a path raises both a `Confirm` and a `Message`, register both: `[HandlerFunctions('ConfirmHandlerYes,PostMessageHandler')]`. When you change the scenario so it no longer hits a dialog, remove that handler from the list. Treat "handler function was not executed" as a signal that the path diverged from what the test claims to exercise, and reconcile the two.
+
+## Anti Pattern
+
+Listing only one of the handlers a path needs — the other UI call goes unhandled and aborts — or padding the list with a handler the path never reaches, which fails the test for the unused handler. Either way the attribute lies about the path, and the failure points at wiring rather than behavior.
```
nikolakukrika: We should update the text a bit with the sample above

microsoft/knowledge/testing/handlers-enqueue-never-assert.good.al
```diff
@@ -0,0 +1,36 @@
...
+    [HandlerFunctions('PostMessageHandler')]
+    procedure PostingShowsConfirmationMessage()
+    var
+        ActualMessage: Text;
+    begin
+        // [WHEN] the code under test posts and raises a Message
+        RunPostingThatMessages();
+
+        // [THEN] the body — not the handler — owns the verdict
+        ActualMessage := LibraryVariableStorage.DequeueText();
```
nikolakukrika: It should be the other way around. Test enqueues the expected values. Handler deques them. 
This way we ensure that we get the UI elements in order with messages we expect. See the comments above about the library.

microsoft/knowledge/testing/handlers-enqueue-never-assert.good.al
```diff
@@ -0,0 +1,36 @@
...
+{
+    Subtype = Test;
+
+    [Test]
+    [HandlerFunctions('PostMessageHandler')]
+    procedure PostingShowsConfirmationMessage()
+    var
+        ActualMessage: Text;
+    begin
+        // [WHEN] the code under test posts and raises a Message
```
nikolakukrika: The test should have an initialize method that clears the library variable storage. Without it if one tests fail to clear it, then all tests will fail.

microsoft/knowledge/testing/handlers-enqueue-never-assert.md
```diff
@@ -0,0 +1,26 @@
...
+
+# Enqueue from handlers; assert in the test body
+
+## Description
+
+A UI handler function runs in its own invocation context, separate from the `[Test]` method's verdict scope. An assertion that fails inside a handler does not reliably surface as the test's failure: the error can be swallowed by the calling UI operation or reported in a way that masks which test failed, so a broken expectation can silently pass. The reliable pattern is to make handlers capture, not judge — push the values they observe into `LibraryVariableStorage.Enqueue` — and let the test body dequeue those values and assert on them, where the verdict belongs. Finishing with `LibraryVariableStorage.AssertEmpty` confirms every expected interaction actually fired and nothing was left unconsumed.
+
+## Best Practice
+
+In the handler, `Enqueue` the message text, the page values, or the confirm question. In the test body, after acting, `Dequeue` each value and verify it with `Assert`; then call `LibraryVariableStorage.AssertEmpty` to prove the handler ran exactly as often as expected. This keeps the pass/fail decision in the method the runner scores and turns a missed or extra UI call into a real failure.
```
nikolakukrika: Should we merge this article with the article above and have a single article about handlers? Both are short.

microsoft/knowledge/testing/handlers-enqueue-never-assert.md
```diff
@@ -0,0 +1,26 @@
...
+technologies: [al]
+countries: [w1]
+application-area: [all]
+---
+
+# Enqueue from handlers; assert in the test body
+
+## Description
+
+A UI handler function runs in its own invocation context, separate from the `[Test]` method's verdict scope. An assertion that fails inside a handler does not reliably surface as the test's failure: the error can be swallowed by the calling UI operation or reported in a way that masks which test failed, so a broken expectation can silently pass. The reliable pattern is to make handlers capture, not judge — push the values they observe into `LibraryVariableStorage.Enqueue` — and let the test body dequeue those values and assert on them, where the verdict belongs. Finishing with `LibraryVariableStorage.AssertEmpty` confirms every expected interaction actually fired and nothing was left unconsumed.
```
nikolakukrika: The article should be the other way around. We should enqueue the expected messages and replies from the test and dequeue them from the handlers. We should verify in the handler if the message is as expected and abort the test early. There needs to be Initialize to clear the variable storage.

microsoft/knowledge/testing/tablerelation-requires-prerequisite-records.bad.al
```diff
@@ -0,0 +1,32 @@
...
+        // before any assertion runs.
+        SalesLine.Init();
+        SalesLine."Document Type" := SalesHeader."Document Type";
+        SalesLine."Document No." := SalesHeader."No.";
+        SalesLine."Line No." := 10000;
+        SalesLine.Type := SalesLine.Type::Item;
+        SalesLine.Validate("No.", 'GHOST');
+        SalesLine.Insert(true);
+
+        Assert.AreEqual('GHOST', SalesLine."No.", 'Unreachable: validation already failed.');
```
nikolakukrika: We should avoid hardcoding the Ghost value inline.

microsoft/knowledge/testing/tablerelation-requires-prerequisite-records.md
```diff
@@ -0,0 +1,26 @@
+---
+bc-version: [all]
+domain: testing
+keywords: [tablerelation, prerequisite, validate, foreign-key, test-data]
+technologies: [al]
+countries: [w1]
+application-area: [all]
+---
+
+# Create related records before inserting test data that points to them
```
nikolakukrika: I wonder if we need this article. This is not very test specific but BC specific. 
I was never thinking of the table relations when writing tests.

microsoft/knowledge/testing/ui-calls-require-test-handlers.md
```diff
@@ -0,0 +1,26 @@
...
+technologies: [al]
+countries: [w1]
+application-area: [all]
+---
+
+# Intercept every UI call with a registered test handler
+
+## Description
+
+Any platform UI interaction the code under test triggers — `Confirm`, `Message`, error dialogs, `Page.Run`/`RunModal`, `Report.Run`/`RunModal`, request pages, `StrMenu`, `Hyperlink`, `Notification.Send` — must be intercepted by a handler function carrying the matching handler attribute and registered on the test method via `[HandlerFunctions]`. A test runs headless: there is no interactive user to dismiss a dialog. If the executed path raises a UI call with no registered handler, the platform throws an "unhandled UI" error and aborts the method. This is a runtime failure, not an assertion failure — the test never reaches its verification, so a reviewer sees an infrastructure error instead of a verdict on the behavior under test.
```
nikolakukrika: There is a lot of overlap on this article and the other articles. I do not like that the samples contain the bad samples as already covered in the previous comments. Can we try to combine everything in one article or maybe 2 articles?

nikolakukrika (changes requested):

microsoft/knowledge/testing/asserterror-needs-expectederror-and-code.bad.al
JesperSchulz: Left the bad sample as-is here: `Customer.Init(); Customer.Name := ''` is the minimal way to get a blank-name record — a Library codeunit would create a *valid* (non-blank) customer and wouldn't exercise the rejection, so the only bad element is the bare `asserterror`. Happy to add an `Initialize()` shell if you'd rather keep the structure symmetric. (373ec31)

microsoft/knowledge/testing/asserterror-needs-expectederror-and-code.good.al
JesperSchulz: Removed the hardcoded string entirely — the sample now calls `Assert.ExpectedTestFieldError(Customer.FieldCaption(Name), '''')`, so there's no literal message left to turn into a label. (373ec31)

microsoft/knowledge/testing/asserterror-needs-expectederror-and-code.good.al
JesperSchulz: Switched to `Assert.ExpectedTestFieldError(...)` as you suggested, so the message and code are no longer hardcoded. Ready to point it at the centralized helper once you've moved it into Library Assert. (373ec31)

microsoft/knowledge/testing/asserterror-needs-expectederror-and-code.md
JesperSchulz: Updated Best Practice to describe the helpers (`ExpectedTestFieldError`, and `ExpectedError`/`ExpectedErrorCode` for other errors) and to say: extend Library Assert (or a local assert library) with a helper rather than hardcode message text and codes across tests. (373ec31)

microsoft/knowledge/testing/handlerfunctions-attribute-must-match-ui-path.good.al
JesperSchulz: Reworked into the merged `ui-handlers-in-tests` article: the test enqueues the expected question and reply, and `ConfirmHandler` dequeues, asserts via `Assert.ExpectedConfirm` (substring), then returns the enqueued reply. `AssertEmpty()` enforces the exact call count. (373ec31)

microsoft/knowledge/testing/handlerfunctions-attribute-must-match-ui-path.good.al
JesperSchulz: Done — handlers now dequeue and verify with `Assert.ExpectedMessage`/`ExpectedConfirm` (substring match) in the merged article. (373ec31)

microsoft/knowledge/testing/handlerfunctions-attribute-must-match-ui-path.md
JesperSchulz: This article was merged into `ui-handlers-in-tests.md`; its Description and Best Practice now describe the enqueue-driven sample. (373ec31)

microsoft/knowledge/testing/handlers-enqueue-never-assert.good.al
JesperSchulz: Reversed as requested: the test enqueues the expected values and replies, and the handlers dequeue + verify. Consolidated into `ui-handlers-in-tests.good.al`. (373ec31)

microsoft/knowledge/testing/handlers-enqueue-never-assert.good.al
JesperSchulz: Added `Initialize()` calling `LibraryVariableStorage.Clear()` at the top of the test, plus `AssertEmpty()` at the end. (373ec31)

microsoft/knowledge/testing/handlers-enqueue-never-assert.md
JesperSchulz: Merged — the three UI/handler articles are now a single `ui-handlers-in-tests` article. (373ec31)

microsoft/knowledge/testing/handlers-enqueue-never-assert.md
JesperSchulz: Rewritten this way in the merged article: expectations are enqueued from the test and verified inside the handlers (substring), with `Initialize()` clearing storage and `AssertEmpty()` confirming counts. Note this does retire the original "never assert in a handler" stance in favor of the library pattern. (373ec31)

microsoft/knowledge/testing/tablerelation-requires-prerequisite-records.bad.al
JesperSchulz: Moot now — this article was dropped; the sample no longer exists. (373ec31)

microsoft/knowledge/testing/tablerelation-requires-prerequisite-records.md
JesperSchulz: Agreed — removed the standalone article. Folded the one test-specific point (a `TableRelation` field is checked on `Validate`/`Insert(true)`, so build parents first) into `use-library-codeunits-for-test-fixtures`. (373ec31)

microsoft/knowledge/testing/ui-calls-require-test-handlers.md
JesperSchulz: Combined the three overlapping handler/UI articles into a single `ui-handlers-in-tests` article, dropping the duplicated bad samples. (373ec31)

JesperSchulz:
@nikolakukrika thanks for the detailed review — pushed 373ec31 addressing all of it. Summary of the rework:

**Consolidation.** The three overlapping handler/UI articles (`ui-calls-require-test-handlers`, `handlerfunctions-attribute-must-match-ui-path`, `handlers-enqueue-never-assert`) are now a single **`ui-handlers-in-tests`** article, so the duplicated Confirm/Message bad samples are gone.

**Handler pattern reversed to the library convention.** The merged sample now has the **test enqueue** the expected text and reply, and the **handlers dequeue + verify** with `Assert.ExpectedConfirm` / `Assert.ExpectedMessage` (substring match, not the full caption) before returning the enqueued reply. Each test starts with an `Initialize()` that calls `LibraryVariableStorage.Clear()` and ends with `AssertEmpty()` to prove the exact interaction count. This intentionally retires the original "never assert in a handler" stance in favor of the library-driven pattern.

**No hardcoded messages/codes.** The `asserterror` sample now uses `Assert.ExpectedTestFieldError(Customer.FieldCaption(Name), '')` instead of a hardcoded message + `'TestField'` code, and the article text points at the Library Assert helpers and recommends extending the library rather than hardcoding. (It assumes `ExpectedTestFieldError` lands in `Library Assert` — happy to align once you've moved it.)

**Dropped the table-relation article.** Agreed it wasn't test-specific enough; its one test-relevant point (a `TableRelation` field is checked on `Validate`/`Insert(true)`, so build parents before children) is folded into `use-library-codeunits-for-test-fixtures`.

**Counts.** 198 → **195** articles. `validate_frontmatter.py`, `Test-KnowledgeIndex.ps1`, and `Build-KnowledgeIndex.ps1` all pass.

I've replied inline on each thread. One open item for you: whether you're comfortable retiring the standalone "never assert in a handler" guidance now that it's expressed through the enqueue/dequeue pattern. Ready for another look when you are.

microsoft/knowledge/testing/asserterror-needs-expectederror-and-code.bad.al
nikolakukrika: Lets keep the structure symmetric. The bad sample should be 99% correct test with only the bad part to highlight. The good tests should contain the correct implementaton

nikolakukrika (approved):

PR #84, opened 2026-07-02
Package BCQuality as an installable plugin (experiment)

gggdttt:
# Package BCQuality as an installable plugin (experiment)

> **Draft / experiment — not for merge as-is.** Opening this to review the packaging approach and gather maintainer feedback on whether consumer glue like this belongs upstream or in a fork.

## Why

We want to evaluate BCQuality as a code-review knowledge base inside [BC-Bench](https://github.com/microsoft/BC-Bench)'s `code-review` category, injected via the cross-CLI plugin mechanism (Copilot CLI **and** Claude Code both support `plugin marketplace add` + `plugin install`). That requires BCQuality to be installable as a plugin. Today it is content only, so a host has no manifest to install and no single entry point to invoke.

## What this adds

Three files, **no changes to knowledge or routing logic**:

| File | Purpose |
|---|---|
| `.claude-plugin/marketplace.json` | Declares a one-plugin marketplace; the plugin `source` is the repo root (`./`) so the whole tree — skills, knowledge, tools — ships with the install and the Entry protocol runs against the installed clone. |
| `.claude-plugin/plugin.json` | The plugin manifest (`name`, `description`, `version`). |
| `skills/bcquality-al-review/SKILL.md` | A thin **bridge skill** in the host-native `SKILL.md` format. Its only job is to bootstrap the existing `skills/entry.md` protocol for a plugin host that knows nothing about BCQuality's layout: refresh the knowledge index (best effort), run Entry with an AL-review task context, follow the dispatch to `microsoft/skills/review/al-code-review.md`, and emit the DO findings contract. |

## Design notes / open questions for maintainers

- **Philosophy tension.** The README states BCQuality "does not contain agents; agents that consume BCQuality ship with orchestrators." The bridge skill *is* consumer glue. Is upstream the right home, or should packaging live in a fork/overlay? This PR is the concrete artifact to discuss that on.
- **Manifest location.** Uses `.claude-plugin/` (Claude convention; also accepted by Copilot CLI in BC-Bench's existing AL-LSP plugin). Copilot docs alternatively allow a root `plugin.json`. Happy to switch if you prefer one.
- **`pwsh` dependency.** The index refresh calls `tools/Build-KnowledgeIndex.ps1`; the skill degrades to path-based discovery when `pwsh` is absent.
- **New top-level `.claude-plugin/`** will trip the advisory `flag-new-top-level` guard — expected, flagging here for visibility.

## Validation

`python .github/scripts/validate_frontmatter.py --root .` → **0 errors, 0 warnings** (the new files fall outside the validator's knowledge/skill scope).

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -1,0 +1,15 @@
+{
+  "name": "bcquality",
+  "owner": {
+    "name": "microsoft/BCQuality",
+    "url": "https://github.com/microsoft/BCQuality"
+  },
+  "plugins": [
+    {
+      "name": "bcquality",
+      "source": "./",
+      "description": "Business Central AL quality knowledge base and review skills, packaged as an installable plugin. Ships the entire BCQuality tree (skills, knowledge, tools) so the Entry routing protocol runs against the installed clone.",
+      "version": "0.1.0"
+    }
+  ]
+}
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,0 +1,9 @@
+{
+  "name": "bcquality",
+  "description": "Quality skills and knowledge for Business Central development. Exposes a review bridge skill that drives the BCQuality Entry protocol over the installed knowledge base.",
+  "version": "0.1.0",
+  "author": {
+    "name": "microsoft/BCQuality",
+    "url": "https://github.com/microsoft/BCQuality"
+  }
+}
--- a/skills/bcquality-al-review/SKILL.md
+++ b/skills/bcquality-al-review/SKILL.md
@@ -1,0 +1,74 @@
+---
+name: bcquality-al-review
+description: Review Business Central AL code changes using the BCQuality knowledge base. Use when reviewing an AL pull request, a working-tree diff, or a single AL file, and you want findings backed by BCQuality's curated, BC-specific quality rules.
+---
+
+# BCQuality AL review
+
+This skill drives the BCQuality **Entry protocol** over the knowledge base that ships
+inside this plugin. It is the plugin entry point for consumers (orchestrators, CLIs)
+that do not already know BCQuality's internal conventions — the only convention they
+need is "invoke this skill for an AL review."
+
+BCQuality itself is orchestrator-agnostic content: knowledge files plus routing and
+action skills. This bridge is the thin consumer glue that lets a plugin host run that
+content without hardcoding BCQuality's layout.
+
+## When to use
+
+- Reviewing an AL pull request or an uncommitted working-tree diff.
+- Reviewing a single AL file.
+- Any task whose goal is "review Business Central / AL code for quality issues."
+
+Do **not** use this skill to *generate* AL code — it only reviews.
+
+## Plugin root
+
+Resolve `PLUGIN_ROOT` to the directory that contains this plugin's
+`.claude-plugin/plugin.json`. This skill lives at
+`PLUGIN_ROOT/skills/bcquality-al-review/SKILL.md`, so `PLUGIN_ROOT` is two levels up
+from this file. All paths below are relative to `PLUGIN_ROOT`. If the host exposes a
+plugin-root environment variable, prefer it.
+
+## Steps
+
+1. **Refresh the knowledge index (best effort).** If `pwsh` is available, run
+   `pwsh PLUGIN_ROOT/tools/Build-KnowledgeIndex.ps1` from `PLUGIN_ROOT` to (re)generate
+   `PLUGIN_ROOT/knowledge-index.json` over the installed tree. This is a discovery
+   accelerator only — if `pwsh` is missing or the build fails, continue; the review
+   skills fall back to path-based discovery.
+
+2. **Run Entry.** Read `PLUGIN_ROOT/skills/entry.md` and execute it against a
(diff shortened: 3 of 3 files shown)


JesperSchulz:
If that would do, I'm ok with it! Not sure why the SKILL is so firm on this only being for review of code, but for a version 1 I guess that's fine.

JesperSchulz:
Overall this is a faithful, well-scoped version of the additive-wrapper approach, and I think it's the right shape: three files, zero changes to knowledge or routing, and a single **bridge skill** that bootstraps the existing Entry -> dispatch -> READ/DO protocol rather than flattening each action skill into a host-native skill. That preserves Entry's routing, super-skill precedence, and the DO output contract, which is the part that's easy to lose when "converting to a plugin" naively. AL-Go and other URL/clone consumers are untouched. Nice.

A few things worth capturing before this moves past experiment:

**1. Upstream vs. fork is the real decision (not a technical blocker).**
The bridge skill is the first client-facing plumbing in a repo that the README defines as pure, orchestrator-agnostic content ("does not contain agents; agents ship with orchestrators"). It isn't an agent, but it *is* consumer glue. The clean alternative is an overlay/fork repo that installs BCQuality and adds only these three files, keeping this repo content-only. This PR is the right artifact to decide that on; I'd lean toward keeping it additive here **only if** we're comfortable that the README's "content only" principle now admits a thin, opt-in host bridge.

**2. Layer pruning / policy has no hook here (known limitation, please note it).**
The bridge always sends `enabled-layers: [microsoft, community, custom]` and ships the whole tree. In the URL/clone model a consumer prunes the clone to policy *before* the agent runs, and the index is then built over the pruned clone. A plugin install has no equivalent per-consumer deny mechanism, so a consumer who wants `/community/` off can't express that through this path today. Worth a sentence in the SKILL notes and/or a follow-up.

**3. Manifest location.** `.claude-plugin/` for cross-CLI is reasonable, but Copilot CLI also accepts a root `plugin.json`. If Copilot is a first-class target, consider dual-homing or documenting the choice so it doesn't silently only work in one host.

**4. Minor:** `PLUGIN_ROOT` resolution ("two levels up / host env var") is a little fragile across hosts, and there are now three places carrying a version (`marketplace.json`, `plugin.json`, per-skill frontmatter) to keep in sync. Fine for an experiment; flagging for when this hardens.

None of these block the experiment. My only ask before merge-as-real (vs. merge-as-experiment) is an explicit maintainer call on #1 and a note on #2.

gggdttt:
> Overall this is a faithful, well-scoped version of the additive-wrapper approach, and I think it's the right shape: three files, zero changes to knowledge or routing, and a single **bridge skill** that bootstraps the existing Entry -> dispatch -> READ/DO protocol rather than flattening each action skill into a host-native skill. That preserves Entry's routing, super-skill precedence, and the DO output contract, which is the part that's easy to lose when "converting to a plugin" naively. AL-Go and other URL/clone consumers are untouched. Nice.
> 
> A few things worth capturing before this moves past experiment:
> 
> **1. Upstream vs. fork is the real decision (not a technical blocker).** The bridge skill is the first client-facing plumbing in a repo that the README defines as pure, orchestrator-agnostic content ("does not contain agents; agents ship with orchestrators"). It isn't an agent, but it _is_ consumer glue. The clean alternative is an overlay/fork repo that installs BCQuality and adds only these three files, keeping this repo content-only. This PR is the right artifact to decide that on; I'd lean toward keeping it additive here **only if** we're comfortable that the README's "content only" principle now admits a thin, opt-in host bridge.
> 
> **2. Layer pruning / policy has no hook here (known limitation, please note it).** The bridge always sends `enabled-layers: [microsoft, community, custom]` and ships the whole tree. In the URL/clone model a consumer prunes the clone to policy _before_ the agent runs, and the index is then built over the pruned clone. A plugin install has no equivalent per-consumer deny mechanism, so a consumer who wants `/community/` off can't express that through this path today. Worth a sentence in the SKILL notes and/or a follow-up.
> 
> **3. Manifest location.** `.claude-plugin/` for cross-CLI is reasonable, but Copilot CLI also accepts a root `plugin.json`. If Copilot is a first-class target, consider dual-homing or documenting the choice so it doesn't silently only work in one host.
> 
> **4. Minor:** `PLUGIN_ROOT` resolution ("two levels up / host env var") is a little fragile across hosts, and there are now three places carrying a version (`marketplace.json`, `plugin.json`, per-skill frontmatter) to keep in sync. Fine for an experiment; flagging for when this hardens.
> 
> None of these block the experiment. My only ask before merge-as-real (vs. merge-as-experiment) is an explicit maintainer call on #1 and a note on #2.

Thanks Jesper!

1 — agreed it's a maintainer call. I lean toward keeping it additive here: since the bridge runs Entry over the whole tree, a fork would need to submodule/copy the tree anyway — more moving parts than a thin opt-in bridge. 

2 — done: enabled-layers now comes from BCQUALITY_ENABLED_LAYERS, and the SKILL notes call out that on a plugin install this only narrows discovery, not a hard deny. Real prune left as a follow-up.

3 — documented in the notes; verified on Copilot CLI. Kept `.claude-plugin/`, noted root plugin.json as dual-home option.

4 — agreed, leaving as-is for the experiment.

Only open item is your call on #1. Thanks!

JesperSchulz (approved):

PR #112, opened 2026-07-15
Add precision guards for systematic agent false-positive patterns

gggdttt:
## Problem

The online eval surfaced several systematic false-positive patterns in the **agent-findings** channel that BC reviewers repeatedly rejected. This PR encodes precision guards so the agent stops emitting them, without suppressing genuine knowledge-backed findings.

## Changes

**`skills/do.md` — shared precision bar** (read by every leaf and the super-skill). Added a "Known false-positive patterns — never emit these" block covering four recurring, reviewer-confirmed FPs:

- **Missing `ToolTip` on a bound page field** — the field inherits the table field's `ToolTip`, and the truly-missing case is already covered by AA0218 (calibrated to `info`). (~27 FP)
- **Page record trigger "missing `exit(true)` blocks the operation"** — the Boolean page triggers (`OnInsertRecord`/`OnModifyRecord`/`OnDeleteRecord`/`OnQueryClosePage`) return `true` by default; only an explicit `exit(false)` cancels the operation. (~7 FP)
- **"Drill-down / lookup / list is unfiltered"** — the effective filter is frequently defined outside the diff (`SourceTableView`, `TableRelation`, `SetRange`/`SetFilter` in unchanged code). (~15 FP)
- **`CalcFields` in both `OnAfterGetRecord` and `OnAfterGetCurrRecord`** — the two triggers fire at different lifecycle points and are not duplicate work. (~11 FP)

**`microsoft/skills/review/al-breaking-changes-review.md` + `al-upgrade-review.md` — released-baseline precondition.** Breaking-change / upgrade findings now require establishing that the affected symbol actually shipped in a released version. Symbols that are new in the app, added-then-changed within the same unreleased cycle, or in an app with no released baseline can be renamed/renumbered/removed freely. Developers rejected ~20 such findings ("app is not live, breaking nobody").

## Notes

- Direction is strictly precision-increasing (makes the agent more conservative); each guard is narrowly scoped to a documented FP pattern with developer confirmation from the eval.
- The page-trigger default-return semantics were confirmed by two BC developers in the eval and match AL Boolean-trigger behavior.
- Suggestion-placement FPs (feedback #4) are engine-side and tracked in a separate BC-ALAgents PR.


skills/do.md
```diff
@@ -158,6 +158,13 @@ When a super-skill rolls up a non-citation finding from a sub-skill (an `id` tha
 - **Never emit** as agent findings: stylistic or formatting preferences (outside a dedicated style skill's own domain); speculative or hypothetical concerns — anything you would phrase with "could", "might", or "consider"; issues that depend on code not visible in the diff; valid alternative approaches; or generic software-engineering advice a competent model already applies without prompting (the same exclusion the knowledge-file admission test enforces).
 - **When in doubt, omit.** Recall is the knowledge files' responsibility; the agent channel exists only for the high-confidence, concrete defect the corpus has not captured yet. A missed low-severity observation is cheaper than a false positive.
 
+**Known false-positive patterns — never emit these.** Recurring agent findings that BC reviewers consistently reject. Each depends on context the diff does not show or on an incorrect model of AL semantics; do not emit them as agent findings from any leaf or super-skill:
```
JesperSchulz: This is too specific for the DO skill. The DO skill is the template every action skill follows. It should not contain specific instructions.

JesperSchulz (changes requested):
You should not change the DO skill with specific instructions for how the review skill should behave.

microsoft/skills/review/al-error-handling-review.md
```diff
@@ -73,6 +73,8 @@ Set `confidence` to:
...
 
+**Known false-positive — do not emit.** The Boolean page triggers `OnInsertRecord`, `OnModifyRecord`, `OnDeleteRecord`, and `OnQueryClosePage` return `true` by default; omitting an explicit return value lets the operation proceed. Only an explicit `exit(false)` (or a path that returns `false`) cancels it. Do not claim a missing `exit(true)` prevents inserts, modifies, or deletes.
```
JesperSchulz: Isn't this knowledge to be tracked in knowledge files? I don't think we should make the skills heavier by adding all these rules. At least that wasn't the initial idea.

JesperSchulz (changes requested):
Knowledge goes into knowledge files, not the skills.

JesperSchulz (approved):

PR #116, opened 2026-07-20
Make BCQuality aware of ToolTips definied on the source object / source field

### LINKED ISSUE #115: Wrong Style Check about ToolTips - ToolTips on Table should be sufficient
For new fields it's totally fine to define toolTips on tables instead of pages.

https://github.com/microsoft/BCQuality/blob/main/microsoft/knowledge/style/tooltip-required-on-page-fields.md

the problem is that this isn't respected.

e.g. here:
https://github.com/microsoft/BCApps/pull/8429#discussion_r3593827632

I don't know if it is intended that non microsoft people change this on the microsoft layer.

pri-kise:
Fixes #115

### DIFF
--- a/microsoft/knowledge/style/tooltip-required-on-page-fields.bad.al
+++ b/microsoft/knowledge/style/tooltip-required-on-page-fields.bad.al
@@ -1,3 +1,15 @@
+tableextension 50251 "Sample Tooltip Bad Ext" extends Customer
+{
+    fields
+    {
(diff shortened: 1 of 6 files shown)


pri-kise:
I abonden this PR in favor of: https://github.com/microsoft/BCQuality/pull/117

that was created without issue -> therefore I didn't know that someone was already working on it.

PR #121, opened 2026-07-24
Authoring-assist: promote to first-class tool with non-blocking PR advisory

dayland:
## Summary

Promotes the **authoring-assist** prototype to a first-class BCQuality feature: it reviews knowledge-article front-matter and proposes missing routing `signals:` (raise triggers + `effect: suppress` suppressors), then posts a **non-blocking, feedback-instrumented advisory comment** on PRs that touch `**/knowledge/**/*.md`.

The advisory is **ADVISORY ONLY** — it never fails a build. Authors apply suggestions via a normal PR under existing `R29` + CI + CODEOWNERS review.

## What's included

| Area | File(s) |
|---|---|
| Suggestion engine (JSON contract, deterministic `suggestionId`, `-ChangedFiles` scoping, **optional** seed) | `tools/Suggest-ArticleSignals.ps1` |
| Advisory comment renderer (upsert anchor + `aa:meta`/`aa:article` markers + reaction footer) | `tools/New-AuthoringAssistComment.ps1` |
| CI: unprivileged intake + trusted runner (review/publish split) + in-repo self-test | `.github/workflows/authoring-assist{,-runner,-selftest}.yml` |
| 23-check smoke test | `tools/Test-SuggestArticleSignals.ps1` |
| Docs | `tools/authoring-assist.md` |
| **Dormant** schema support: `R29` validates the optional `signals` block shape | `.github/scripts/validate_frontmatter.py` |

## Dormant schema — safe by design

The `signals`/`effect` keys are **present but dormant**. No production pipeline consumes them:
- The PR Reviewer only reads `effect: suppress` behind its `BCQ_INDEX_V2` flag (off in prod) and falls back safely when absent.
- The reviewer never invokes `validate_frontmatter.py`.
- Existing rules (incl. `R24` skill-id uniqueness, `R27`, `R28`) are unchanged; `R29` is purely additive.

## Explicitly out of scope

- The **routing-index tuning thread** (generator/seed/index schema) — separate work. The tool runs self-contained without it and auto-enriches if a seed later lands.
- **Harvest + aggregation** of acceptance/reactions — a separate BC-ALAgentsInternal PR (PR 2).

## Validation

- Front-matter validator: **0 errors** (R29 active)
- Smoke test: **23/23** (seed-free graceful degradation covered)
- All workflows YAML-valid

## Rollout note for reviewers

Draft pending: confirm workflow permissions posture (unprivileged `pull_request` intake + trusted `workflow_run` publish) and whether to pin actions to SHAs (repo currently uses `@vN` tags).


JesperSchulz (changes requested):
Two blockers before this is safe and reliable to merge:

- `.github/workflows/authoring-assist-runner.yml:67`: `workflow_run.head_branch` is controlled by a fork PR and is interpolated directly into a trusted PowerShell script. A branch name containing a quote plus PowerShell syntax can escape this assignment and execute with the runner job token. Pass event values through `env:` and read them from `$env:` instead of embedding `${{ ... }}` in the script.
- `.github/workflows/authoring-assist.yml:17` and the runner's zero-count conditions: when knowledge changes are deleted or fully reverted, the path filter and `count != 0` skips prevent publishing an empty result, so an advisory from an older head can remain visible and actionable. Ensure synchronize cleanup still runs and emits/publishes an empty report so the existing comment is updated or removed; the publish job already has the intended `articleCount == 0` behavior.

PR #125, opened 2026-08-05
knowledge(data-modeling): TableRelation delete/rename asymmetry and the xRec before-image contract

StefanMaron:
Three related `data-modeling` concerns, each backed by executable tests rather than by memory. They came from a real defect that shipped through a six-reviewer panel without being caught — which shows these tests catch a case that actually happened in production, not just a hypothetical one.

### `owning-table-must-delete-dependents-in-ondelete`

AL has no cascading delete. This is easy to miss because of one inconsistency: the platform *does* keep references correct on rename via `TableRelation`, so a developer who sees that and assumes the same is true for delete ends up leaving orphan records behind. This also covers a permission issue: delete rights are needed on the dependent table, not just the parent.

### `validate-table-relation-false-suppresses-rename-propagation`

The less obvious one. The property name suggests it only affects input validation, but turning it off also turns off rename propagation. Verified against a parent renamed once while a child held three fields: a normal relation (updates), the same relation with `ValidateTableRelation = false` (does **not** update), and a field with no relation at all (does not update) — the third case is the control that proves the test can detect a field that fails to update.

### `xrec-is-a-before-image-only-in-some-triggers`

Corrects two common but incorrect assumptions: that `xRec` is always the previous record, and that it "only works from a page." The behavior depends on the trigger: a true before-image in `OnRename` and `OnDelete` regardless of what triggered the write; a copy of `Rec` in `OnInsert`/`OnModify` when triggered from code; a true before-image in those same two triggers only when a page triggered the write. That difference is why an `OnModify` comparison against `xRec` passes manual page testing but silently does nothing in a job queue.

## Provenance

The behavioral claims are proven by tests in a public AL-language conformance suite rather than asserted from experience:

- `xRec` per-trigger behavior, code-driven — merged and passing.
- `xRec` page-driven behavior, and the `TableRelation` / `ValidateTableRelation` propagation matrix — same suite.

Test values are chosen so that old, new, and type-default values are all different from each other, so "`xRec` was never populated" shows up as its own distinguishable result instead of being lumped in with "not the old record."

## Checks

- Targets `/community` per CONTRIBUTING — general BC knowledge, not specific to this fork.
- Frontmatter validator: 0 errors, 0 warnings.
- 36–38 lines each (well under the 100-line limit); one concern per file; no fenced code blocks — examples are sibling `.good.al` / `.bad.al` files.
- `pwsh ./tools/Test-ReviewFixtures.ps1 -Root .` passes — 32 cases, 16 leaf domains.

### DIFF
--- a/community/knowledge/data-modeling/owning-table-must-delete-dependents-in-ondelete.bad.al
+++ b/community/knowledge/data-modeling/owning-table-must-delete-dependents-in-ondelete.bad.al
@@ -1,0 +1,48 @@
+table 50100 "Order Header"
+{
+    fields
+    {
+        field(1; "Entry No."; Integer)
+        {
+            DataClassification = CustomerContent;
+        }
+    }
+
+    keys
+    {
+        key(PK; "Entry No.")
+        {
+            Clustered = true;
+        }
+    }
+
+    // No OnDelete. Deleting a header silently orphans every Order Line that
+    // belonged to it. Nothing errors, and no page shows the stranded rows.
+}
+
+table 50101 "Order Line"
+{
+    fields
+    {
+        field(1; "Line No."; Integer)
+        {
+            DataClassification = CustomerContent;
+        }
+        field(2; "Header Entry No."; Integer)
+        {
+            DataClassification = CustomerContent;
+            // Reads like referential integrity. It is lookup and input validation
+            // only: it propagates a RENAME of the parent key, and cascades nothing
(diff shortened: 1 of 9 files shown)


JesperSchulz (approved):

JesperSchulz:
Let's keep those in the community layer for a short while. I actually think those should be promoted to the Microsoft layer!

PR #130, opened 2026-08-17
knowledge(performance): align SetLoadFields placement with AL Guidelines

### LINKED ISSUE #120: Clarify the correct order of Filtering and SetLoadFields
## Description

There appears to be conflicting guidance regarding the recommended order of `SetRange` / `SetFilter` and `SetLoadFields`.

The AL Guidelines explicitly mark the following order as bad code:

```al
Item.SetLoadFields("Item Category Code");
Item.SetRange("Third Party Item Exists", false);
Item.FindFirst();
```

They recommend applying the filter first:

```al
Item.SetRange("Third Party Item Exists", false);
Item.SetLoadFields("Item Category Code");
Item.FindFirst();
```

Source:

https://github.com/microsoft/alguidelines/blob/main/content/docs/BestPractices/SetLoadFields/Index.md#bad-code-1

However, the BCQuality good example uses the opposite order:

```al
Customer.SetLoadFields(Name);
Customer.SetRange("Country/Region Code", 'US');
if Customer.FindSet() then
```

Source:

https://github.com/microsoft/BCQuality/blob/8fb7f61808c5ddacc809abb5aac5230db1186cca/microsoft/knowledge/performance/use-setloadfields-for-partial-records.good.al#L7-L16

Microsoft Learn states that `SetLoadFields` configures the fields for subsequent record loads and that fields used for filtering are loaded automatically. It does not appear to specify a required order between `SetRange` and `SetLoadFields`.

## Question

Which pattern should BCQuality document?

Should the BCQuality example be changed to match the AL Guidelines?
Does BCQuality intentionally recommend a different order?
If both variants are valid, should that be stated explicitly to avoid contradictory guidance?

## Expected outcome

The repositories should provide consistent guidance and clearly state whether the order is technically relevant or only a coding convention.

JesperSchulz:
Resolves the contradiction reported in #120.

## The question

The [AL Guidelines](https://github.com/microsoft/alguidelines/blob/main/content/docs/BestPractices/SetLoadFields/Index.md#bad-code-1) mark `SetLoadFields` placed *before* `SetRange`/`SetFilter` as bad code; BCQuality's samples used exactly that order.

## The answer

**The order is a readability convention, not a behavioural one.** Per [`Record.SetLoadFields`](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/methods-auto/record/record-setloadfields-method) on Learn:

> Depending on the runtime version, the runtime may require extra fields to be selected for loading. […] For example, **fields that are filtered upon are always loaded**.

So both orders materialize an identical projection. `SetRange`/`SetFilter` never resets the load-field selection — only a fieldless `SetLoadFields()` (which resets to all readable normal fields) or a later `SetLoadFields(...)` overwriting an earlier one changes what is loaded.

The upstream rule is about keeping `SetLoadFields` adjacent to the read it governs — the same rationale as the AL Guidelines' case-statement example, which BCQuality already mirrors in `load-common-fields-before-branching-on-case.md`. That page is tagged `Readability`, not `Performance`.

## Changes

BCQuality now follows the AL Guidelines order, and states explicitly that the order is not technically relevant so the two repos cannot be read as contradicting each other.

- Reordered filters ahead of `SetLoadFields` in the six affected AL samples:
  - `use-setloadfields-for-partial-records.{good,bad}.al`
  - `load-only-primary-key-fields-for-reference-work.good.al`
  - `avoid-cloning-records-before-modify-delete-in-loops.{good,bad}.al`
  - `use-setautocalcfields-for-per-row-flowfields.{good,bad}.al`
- `use-setloadfields-for-partial-records.md`:
  - **Best Practice** — place the call immediately before the read, after any `SetRange`/`SetFilter`.
  - **Description** — the order does not change the projection; only a fieldless or overwriting `SetLoadFields` does.
  - **Anti Pattern** — statement order is explicitly *not* part of the anti pattern, so a review agent reports the reverse order as a readability observation at most, never as a performance defect.

The Anti Pattern placement is deliberate: per `skills/read.md`, only `## Best Practice` and `## Anti Pattern` are normative, so a false-positive suppression has to live there to be honoured by consumers.

## Validation

All three CI validators pass locally:

- `validate_frontmatter.py` — 0 errors, 0 warnings
- `Test-KnowledgeIndex.ps1` — 255 articles, deterministic, full coverage
- `Test-ReviewFixtures.ps1` — 32 cases across 16 leaf domains

Closes #120


microsoft/knowledge/performance/avoid-cloning-records-before-modify-delete-in-loops.good.al
```diff
@@ -4,8 +4,8 @@ codeunit 50492 "Perf Record Clone Good"
     var
         Customer: Record Customer;
     begin
-        Customer.SetLoadFields("Credit Limit (LCY)");
         Customer.SetFilter("Credit Limit (LCY)", '>0');
+        Customer.SetLoadFields("Credit Limit (LCY)");
```
BardurKnudsen, line 8: How is this different than in the 'before' / 'bad' code?

microsoft/knowledge/performance/avoid-cloning-records-before-modify-delete-in-loops.bad.al
```diff
@@ -5,8 +5,8 @@ codeunit 50493 "Perf Record Clone Bad"
         Customer: Record Customer;
         CustomerCopy: Record Customer;
     begin
-        Customer.SetLoadFields("Credit Limit (LCY)");
         Customer.SetFilter("Credit Limit (LCY)", '>0');
+        Customer.SetLoadFields("Credit Limit (LCY)");
```
BardurKnudsen, line 9: Why is this bad? We are just setting some parameters on the record variable before issuing the sql command (FindSet() -> Select).
If anything, you may say it's not necessary to specify a field that is being filtered on. But even that is harmless.

microsoft/knowledge/performance/load-only-primary-key-fields-for-reference-work.good.al
```diff
@@ -6,8 +6,8 @@ codeunit 50100 "Item Reindex Queue"
         ReindexQueue: Codeunit "Reindex Queue";
     begin
         // Only the primary key is used in the loop body; load nothing else.
-        Item.SetLoadFields("No.");
         Item.SetRange("Item Category Code", CategoryCode);
+        Item.SetLoadFields("No.");
```
BardurKnudsen, line 10: Why is this better?

BardurKnudsen (approved):

microsoft/knowledge/performance/load-only-primary-key-fields-for-reference-work.good.al
JesperSchulz: It is not better at runtime; both orders produce the same projection because filtered fields are included at read time. Filters-first is only the AL Guidelines readability convention: keeping `SetLoadFields` adjacent to the read makes its scope easier to see. The article now makes that distinction explicit so agents do not report the opposite order as a performance issue.

microsoft/knowledge/performance/avoid-cloning-records-before-modify-delete-in-loops.good.al
JesperSchulz: The `SetLoadFields` ordering is intentionally identical in the good and bad companions because it is not what distinguishes them. This good sample modifies the iterating `Customer` record directly; the bad sample clones it into `CustomerCopy` before modifying, which restarts the SQL query per record. I reordered the incidental setup in both samples only to align with the AL Guidelines readability convention.

microsoft/knowledge/performance/avoid-cloning-records-before-modify-delete-in-loops.bad.al
JesperSchulz: Agreed: this `SetFilter`/`SetLoadFields` sequence is harmless and is not why the file is the bad companion. The anti-pattern is `CustomerCopy.Copy(Customer)` followed by modifying the clone, which causes an extra SQL statement per row. The ordering change merely keeps all samples consistent with the AL Guidelines convention; the article explicitly says the reverse order is not a performance defect.

BardurKnudsen (approved):

PR #132, opened 2026-08-18
Refine self-improvement review guidance

gggdttt:
## Summary

Refine the second self-improvement batch into evidence-backed BCQuality guidance rather than copying reaction-generated changes verbatim.

- narrow `IsHandled` reset findings to values that can actually carry `true`, including loop-carried state
- retire local Label scope as a correctness/localization rule
- allow semantic capture/reset/assert UI-handler tests and reserve queues for interaction contracts
- require restart-safe persisted checkpoints without weakening bounded retrieval
- align `ModifyAll`/`DeleteAll` fallback guidance with documented platform conditions

## False-positive boundaries

- A fresh local Boolean is already `false` for one non-looping raise. A local declared outside a loop initializes once and can carry `true` across iterations. A later raise reached only after a semantically valid `if IsHandled then exit;` remains proven false.
- Independent event seams use independent handled state so one handled event does not suppress a downstream publisher.
- Procedure-local Labels are supported and included in the extension translation pipeline; object scope is only a possible maintainability convention.
- `[HandlerFunctions]` proves listed handlers execute. Missing semantic verification is `minor`; queue storage is not mandatory for every handler. Fixtures cover omitted handlers, listed-but-unreached handlers, and meaningful capture/assert verification.
- Persisted retry progress and bounded next-N retrieval are separate requirements. A full-tail `FindSet` remains unbounded, and the checkpoint sample initializes its persisted watermark on cold start.
- Progress UX permits an explicit row loop only when the equivalent bulk API already falls back to individual operations and semantics are preserved. DeleteAll guidance includes trigger code, related subscribers, security filtering, Media/MediaSet, and companion-field fallback conditions.

## Validation

- `python .github/scripts/validate_frontmatter.py --root .`
- `tools/Test-ReviewFixtures.ps1 -Root . -PrepareDirectory <temp>`
- `.github/scripts/Test-KnowledgeIndex.ps1 -Root .`
- `git diff --check`

## Deliberately unchanged

- `internal-access-is-not-a-security-boundary`
- ErrorInfo boundary guidance
- the PR #7938 unreachable-case candidate
- no swallowed-handler-assertion warning
- existing article slugs and paths


gggdttt:
Gentle ping — this has been open since 18 August with all checks green and no review yet.

Context on why it is shaped the way it is, in case the size is what is holding it up. The substance is five knowledge articles, corrected after adjudicating false positives from the PR-review self-improvement runs on BCApps 9067, 7938 and 8553:

- `events/initialize-ishandled-to-false-before-publishing` — tighten the reset requirement to the case that actually needs it. A fresh local Boolean is initialised once per procedure invocation, so a non-looping single raise does not need an explicit reset; a raise inside a loop does, because the value is carried across iterations. Also stops the `.good.al` sample from using one broad `exit` that suppresses independent downstream event seams.
- `style/labels-declared-at-object-scope` — drop the blanket rule and the two samples. A procedure-local `Label` is legitimate, so the previous wording produced false positives.
- `testing/ui-handlers-in-tests` — allow a handler to capture meaningful state and assert on it after the run, instead of requiring it to be empty.
- `performance/avoid-commit-inside-loops` and the two `modifyall` articles — separate restart-safe checkpointing from bounded retrieval, and limit the bulk-operation progress-loop exemption.

The eight `microsoft/skills/review/*.md` edits are one-line cue updates that follow from those articles, and the single `skills/do.md` line just removes "moving a local `Label` to object scope" from the mechanical-fix examples, since that is no longer a finding after the `labels-declared-at-object-scope` change.

If the `skills/do.md` line is the blocker because that file needs both maintainers, I am happy to split it out so the knowledge and skill-cue changes can land first. This PR is currently blocking one further candidate from a later run, so an unblock in either direction is useful.

microsoft/knowledge/events/initialize-ishandled-to-false-before-publishing.good.al
gggdttt, line 1: The tile of this article should be updated to fit the content

JesperSchulz (changes requested):
One blocking false-positive boundary remains in the UI-handler guidance.

`microsoft/knowledge/testing/ui-handlers-in-tests.md` now says every listed handler must execute, and `microsoft/skills/review/al-testing-review.md` requires `[HandlerFunctions]` to exactly match invoked handlers. The linked Microsoft documentation is narrower: every **nonoptional** handler must execute, and it explicitly permits send-notification and recall-notification handlers to be optional. As written, agents can flag or remove a deliberately unused optional notification handler.

Please qualify the article's two-sided-contract wording and the review cue to exempt optional `SendNotificationHandler`/`RecallNotificationHandler` entries (and ideally encode that boundary in the fixture). The rest of the event-state, Label-scope, checkpoint, bulk-operation, and semantic UI-verification changes look sound.

gggdttt:
Thanks, you caught a real bug. Fixed in c213f14.

You were right that the article was too strict, but the rule turned out to be a bit narrower than "notification handlers are exempt", so I want to explain what I did.

**Optional is something you declare, not something the handler type gives you.** The attribute takes a boolean: `[SendNotificationHandler([HandlerIsOptional: Boolean])]`, same for `RecallNotificationHandler`. So `[SendNotificationHandler(true)]` is optional, but `[SendNotificationHandler]` written plain is nonoptional like any other handler. If I had exempted all notification handlers, I would have traded a false positive for a false negative.

The docs never say what the default is when you leave the argument out, so I counted the real usage in `microsoft/BCApps` on `origin/main`:

| form | count |
|---|---|
| `[SendNotificationHandler(true)]` | 10 |
| `[SendNotificationHandler(false)]` | 0 |
| `[SendNotificationHandler]` | 359 |
| `[RecallNotificationHandler(true)]` / `(false)` | 0 / 0 |
| `[RecallNotificationHandler]` | 103 |

Ten people wrote `(true)` on purpose and nobody ever wrote `(false)`. That only makes sense if the plain form is nonoptional, otherwise those ten annotations do nothing. So the `.bad.al` case is now the same test as `.good.al` with only that one argument removed.

**The article had a second way to produce this false positive.** I built a small synthetic PR with a `[SendNotificationHandler(true)]` that the test path never triggers, then ran the engine twice per side and changed nothing but the BCQuality tree:

| round | BCQuality | findings | testing leaf |
|---|---|---|---|
| 1 | `origin/main` | 1 blocker | worklist=2, 1 finding |
| 2 | `origin/main` | 1 blocker | worklist=2, 1 finding |
| 1 | this PR | 0 | worklist=2, 0 findings |
| 2 | this PR | 0 | worklist=2, 0 findings |

Both baseline runs flagged the handler, but not for the reason you expected. Neither one complained that it never executed. They complained it had no enqueue/dequeue and no `LibraryVariableStorage.AssertEmpty`, because the old Best Practice made those mandatory for every handler. Asking a test to assert an empty queue for a notification that may never fire does not make sense. The article now asks for queue storage only when order, count, text, replies, or a scripted sequence is actually part of the contract, and the Anti Pattern says not to flag its absence on its own.

Worth noting: in both candidate runs the testing leaf still loaded and evaluated the article (worklist=2, evaluated=2) and returned nothing. So it is suppressed on purpose, not just missed during retrieval.

**The part that worried me more was in the skill, not the article.** `al-testing-review.md` had "remove a handler name in `HandlerFunctions`" in its mechanical-fix list, and under the `do.md` contract that renders as a one-click suggestion. So the agent would not just flag a deliberately optional notification handler, it would hand the developer a button to delete it. One click, and the test breaks on every run where the notification does fire. That entry is now excluded for optional notification handlers. On top of that, the token list used for relevance had no `SendNotificationHandler` or `RecallNotificationHandler` in it at all, so the skill could not even see them. Both are added.

**Fixture.** `evaluation/review-fixtures.json` now pins the `testing` domain to `ui-handlers-in-tests`, so this pair is the domain's control. `.good.al` has a listed but never raised `[SendNotificationHandler(true)]` and has to stay clean. `.bad.al` has the plain nonoptional form on a path that is never reached and has to be flagged. Neither sample has explanatory comments, so the clean control stays honest.

One thing I should flag: `[RecallNotificationHandler(true)]` has zero precedent in BCApps. The docs give it the same signature as `SendNotificationHandler`, so I covered it the same way, but that half is based on the documentation rather than on code anyone has actually written.

JesperSchulz (approved):
The update resolves the optional-handler false positive precisely.

- Optionality is keyed to the explicit `HandlerIsOptional` argument, so `[SendNotificationHandler(true)]` and `[RecallNotificationHandler(true)]` may remain listed without executing, while the plain nonoptional forms still must execute.
- The reviewer no longer proposes deleting optional notification handlers.
- Relevance tokens and the paired good/bad fixture now exercise the exact boundary.

The rest of the previously reviewed guidance remains sound.

BardurKnudsen (approved):

microsoft/knowledge/events/reset-ishandled-only-when-the-value-can-carry-over.md
```diff
@@ -0,0 +1,26 @@
...
+application-area: [all]
+---
+
+# Reset IsHandled before publishing only when its value can carry over
+
+## Description
+
+A routine that raises an `OnBefore…` integration event with a `var IsHandled: Boolean` parameter passes that variable by reference, so a pre-existing `true` can affect the following control flow. AL [automatically initializes Boolean variables to `false`](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-al-variables#initialization), so a freshly declared local Boolean passed to one event exactly once per procedure invocation is already deterministic. Initialization does not repeat for each loop iteration: a local declared outside a loop can carry `true` from one iteration to the next even when the source contains only one textual event raise. Outside a loop, reaching a later raise after `if IsHandled then exit;` also proves the value is `false`, provided that early exit is semantically correct and does not skip required downstream events.
+
+## Best Practice
```
nikolakukrika, line 16: IsHandled is bad in general. You can check with Marko Aleksandric to update these rules. I would state DO NOT INTRODUCE IsHandled :)

microsoft/knowledge/performance/avoid-commit-inside-loops.md
```diff
@@ -13,16 +13,18 @@ application-area: [all]
 
 ## Description
 
-Commit ends the current write transaction. Calling it inside a per-row loop produces one transaction per iteration and loses the ability to roll back the whole operation atomically; it also interferes with the platform's ability to batch write operations. Most loops need no explicit Commit at all — AL auto-commits the enclosing code module on successful completion (see `understand-implicit-transaction-boundary.md`). When the batch is too large for one transaction, the fix is not a per-row Commit but bounded checkpoints that select an exact list of at most N keys and process only those rows.
+Commit ends the current write transaction. Calling it inside a per-row loop usually produces one transaction per iteration and loses the ability to roll back the whole operation atomically; it also interferes with batching. Most loops need no explicit Commit at all — AL auto-commits the enclosing code module on successful completion (see `understand-implicit-transaction-boundary.md`).
```
nikolakukrika, line 16: Actually this can be a pattern. We commit ever x rows, or every x minutes. We need to ensure that the loop is possible to be entered again.

microsoft/knowledge/performance/avoid-commit-inside-loops.bad.al
```diff
@@ -3,12 +3,24 @@ codeunit 50129 "Perf Sample CommitInLoop Bad"
...
+
+                // LastCustomerNo exists only in memory, so a retry cannot exclude
+                // work that was already committed.
+                LastCustomerNo := Customer."No.";
+                ProcessedCount += 1;
+
+                // This still opened a FindSet over the complete remaining tail;
+                // periodic commits do not turn retrieval into bounded TOP X.
+                if ProcessedCount mod 500 = 0 then
+                    Commit();
```
nikolakukrika, line 23: Shouldn't we memorize where we have stopped somehow? LastCustomerNo is not tracked anywhere. If this one fails, we will go through the list all over again. 

There should be a sample with time too.

microsoft/knowledge/performance/avoid-commit-inside-loops.bad.al
```diff
@@ -3,12 +3,24 @@ codeunit 50129 "Perf Sample CommitInLoop Bad"
...
     begin
+        Customer.SetFilter("No.", '>%1', LastCustomerNo);
         if Customer.FindSet(true) then
             repeat
                 Customer.Name := UpperCase(Customer.Name);
                 Customer.Modify();
-                Commit();
+
+                // LastCustomerNo exists only in memory, so a retry cannot exclude
+                // work that was already committed.
```
nikolakukrika, line 16: This one is OK, we only waste time. It is safe to Rerun, there is no partial commits etc...

microsoft/knowledge/performance/avoid-commit-inside-loops.good.al
```diff
@@ -16,11 +16,22 @@ codeunit 50128 "Perf Sample CommitInLoop Good"
...
-        // The outer loop owns checkpoints; the per-row loop contains no Commit.
-        while NormalizeNextChunk(LastCustomerNo) do
+        if not NormalizeState.Get('CUSTOMER') then begin
+            NormalizeState.Init();
+            NormalizeState.Code := 'CUSTOMER';
+            NormalizeState.Insert();
+        end;
+        LastCustomerNo := NormalizeState."Last Customer No.";
+
+        while NormalizeNextChunk(LastCustomerNo) do begin
```
nikolakukrika, line 29: It would be good to have the bad sample and good sample at the same time that is using the same code.

microsoft/knowledge/testing/ui-handlers-in-tests.good.al
```diff
@@ -1,57 +1,49 @@
...
-    [ConfirmHandler]
-    procedure ConfirmHandler(Question: Text[1024]; var Reply: Boolean)
+    [ModalPageHandler]
+    procedure CustomerCardHandler(var CustomerCard: TestPage "Customer Card")
     begin
-        // Verify the RIGHT dialog fired (substring match), then return the
-        // reply the test enqueued for it.
-        Assert.ExpectedConfirm(LibraryVariableStorage.DequeueText(), Question);
-        Reply := LibraryVariableStorage.DequeueBoolean();
+        CapturedCustomerNo := CustomerCard."No.".Value();
```
nikolakukrika, line 36: We should use library variable storage here. First to ensure that the ConfirmHandler is expected.
Then to dequeue the reply and store the value. This way we ensure that we do not dismiss the dialogs blindly, that we are getting exact number of dialogs as expected.

microsoft/knowledge/testing/ui-handlers-in-tests.good.al
```diff
@@ -1,57 +1,49 @@
...
-    begin
-        // Stands in for the production routine that confirms, then messages.
-        if Confirm('Post this document?', false) then
-            Message('Posting completed.');
+        Page.RunModal(Page::"Customer Card", Customer);
+
+        Assert.AreEqual(Customer."No.", CapturedCustomerNo, 'The customer card opened for the wrong customer.');
     end;
 
-    [ConfirmHandler]
```
nikolakukrika, line 39: Ideally we should get the confirm handler back to have good and bad scenarios aligned

microsoft/knowledge/testing/ui-handlers-in-tests.md
```diff
@@ -1,28 +1,30 @@
...
+List the handlers the scenario triggers, keep an optional notification handler listed for a notification the scenario may conditionally raise, and make each executed handler contribute meaningful evidence. For a single modal page, reset a capture variable before the action, capture a concrete value from the page in the handler, and assert the expected value after `RunModal`. For ordered or repeated interactions, let the test enqueue expectations, let handlers dequeue and verify them, clear storage during initialization, and finish with `AssertEmpty`.
```
nikolakukrika, line 22: The text about using library variable storrage should be listed.
There should be a strong recommendation to have one handler of a given type if possible

microsoft/skills/review/al-events-review.md
```diff
@@ -51,7 +51,7 @@ When the post-conflict worklist is empty because no applicable events knowledge
 
 The following targeted checks map diff signals to specific `events` articles. Treat each as a candidate-selection cue: when the signal appears in the changed code, add the named article to the worklist and evaluate it in Action.
 
-- `IsHandled` raised without an immediately preceding `IsHandled := false;`, or one `IsHandled` variable reused across several raises with no reset between them — `initialize-ishandled-to-false-before-publishing`.
+- An `IsHandled` value that can carry over as `true` (reused after an earlier raise, re-entered on a later loop iteration, input/global/field, or otherwise seeded) is passed to a publisher without a reset — `reset-ishandled-only-when-the-value-can-carry-over`. Do not match one non-looping raise using a fresh local Boolean, or a later raise reached only after a semantically valid `if IsHandled then exit;` proves the value is false.
```
nikolakukrika, line 54: Is handled is bad, sync with Marko

PR #133, opened 2026-08-19
Add TransferFields SkipFieldsNotMatchingType guidance

yahyatouil-dev:
## What

One knowledge rule: don't use `TransferFields(..., SkipFieldsNotMatchingType: true)`
as a blanket way to keep two evolving tables transferring without
errors, when the destination actually depends on every source field
being copied.

## Why this belongs in BCQuality

`TransferFields` only errors on a type mismatch when both tables are
in the same extension - cross-extension mismatches are already
skipped by default, flag or not. That distinction isn't obvious from
the method signature, and it's exactly the kind of thing where an AI
(or a dev in a hurry) reaches for `SkipFieldsNotMatchingType = true`
to make a compile/runtime error go away without noticing a field
silently stopped being copied.

## Files

- `community/knowledge/data-modeling/transferfields-skip-type-mismatch-can-drop-data.md`
- `.good.al` - explicit field-by-field mapping with `Evaluate` for the
  type conversion, so a bad value fails loudly instead of vanishing
- `.bad.al` - two tables in the same app, `TransferFields(Source, true, true)`,
  `Reference` silently dropped because it's `Code[20]` on one side and
  `Integer` on the other

## Verified against

- Record.TransferFields(var Record, Boolean, Boolean) - Microsoft Learn
- Confirmed the same-extension vs. cross-extension error behavior
  independently against community testing before writing the rule

One concern, one file, community layer.

### DIFF
--- a/community/knowledge/data-modeling/transferfields-skip-type-mismatch-can-drop-data.good.al
+++ b/community/knowledge/data-modeling/transferfields-skip-type-mismatch-can-drop-data.good.al
@@ -1,0 +1,56 @@
+table 50121 "Transfer Source"
+{
+    fields
+    {
+        field(1; "Entry No."; Integer)
+        {
+            DataClassification = CustomerContent;
+        }
+        field(2; "Reference"; Code[20])
+        {
+            DataClassification = CustomerContent;
+        }
+    }
+
+    keys
+    {
+        key(PK; "Entry No.")
+        {
+            Clustered = true;
+        }
+    }
+
+}
+
+table 50122 "Transfer Target"
+{
+    fields
+    {
+        field(1; "Entry No."; Integer)
+        {
+            DataClassification = CustomerContent;
+        }
+        field(2; "Reference"; Integer)
+        {
+            DataClassification = CustomerContent;
+        }
+    }
+
+    keys
+    {
+        key(PK; "Entry No.")
+        {
+            Clustered = true;
+        }
+    }
+
+}
+
+codeunit 50491 "TransferFields Good"
+{
+    procedure CopyData(Source: Record "Transfer Source"; var Target: Record "Transfer Target")
+    begin
+        Target."Entry No." := Source."Entry No.";
+        Evaluate(Target."Reference", Source."Reference");
+    end;
+}
--- a/community/knowledge/data-modeling/transferfields-skip-type-mismatch-can-drop-data.good.al
+++ b/community/knowledge/data-modeling/transferfields-skip-type-mismatch-can-drop-data.good.al
@@ -49,8 +49,11 @@ table 50122 "Transfer Target"
 codeunit 50491 "TransferFields Good"
 {
     procedure CopyData(Source: Record "Transfer Source"; var Target: Record "Transfer Target")
+    var
+        ConvertedReference: Integer;
     begin
         Target."Entry No." := Source."Entry No.";
-        Evaluate(Target."Reference", Source."Reference");
+        Evaluate(ConvertedReference, Source."Reference");
+        Target.Validate("Reference", ConvertedReference);
     end;
 }
--- a/community/knowledge/data-modeling/transferfields-skip-type-mismatch-can-drop-data.md
+++ b/community/knowledge/data-modeling/transferfields-skip-type-mismatch-can-drop-data.md
@@ -1,0 +1,26 @@
+---
+bc-version: [all]
+domain: data-modeling
+keywords: [transferfields, skipfieldsnotmatchingtype, type-mismatch, field-mapping, data-transfer]
+technologies: [al]
+countries: [w1]
+application-area: [all]
+---
+
+# Do not use SkipFieldsNotMatchingType to hide required TransferFields mismatches
+
+## Description
+
+`Record.TransferFields` copies values between fields with matching field numbers. Without `SkipFieldsNotMatchingType` (or with it `false`), a type mismatch between two fields in the same extension raises a runtime error at the point of transfer. Setting `SkipFieldsNotMatchingType` to `true` removes that error: the field is skipped instead, and the rest of the transfer completes normally. The caller gets no indication that a field was not copied.
+
+## Best Practice
+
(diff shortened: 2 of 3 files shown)


yahyatouil-dev:
@microsoft-github-policy-service agree

community/knowledge/data-modeling/transferfields-skip-type-mismatch-can-drop-data.md
```diff
@@ -0,0 +1,26 @@
...
+
+# Do not use SkipFieldsNotMatchingType to hide required TransferFields mismatches
+
+## Description
+
+`Record.TransferFields` copies values between fields with matching field numbers. Without `SkipFieldsNotMatchingType` (or with it `false`), a type mismatch between two fields in the same extension raises a runtime error at the point of transfer. Setting `SkipFieldsNotMatchingType` to `true` removes that error: the field is skipped instead, and the rest of the transfer completes normally. The caller gets no indication that a field was not copied.
+
+## Best Practice
+
+Use `TransferFields(Source)` when matching field definitions are an expected part of the table design. If source and destination fields intentionally have different types, map those fields explicitly and handle the conversion or validation in code. Use `SkipFieldsNotMatchingType = true` only when skipping incompatible fields is an intentional, documented part of the transfer contract.
```
JesperSchulz: The one-argument overload is not a complete fail-fast replacement here. It defaults `InitPrimaryKeyFields` to `true`, silently ignores destination fields with no matching source field number, and cross-extension type mismatches are skipped even when `SkipFieldsNotMatchingType` is `false`. Please preserve the intended primary-key behavior and state that required fields, especially cross-extension fields, must be mapped or verified explicitly.

community/knowledge/data-modeling/transferfields-skip-type-mismatch-can-drop-data.md
```diff
@@ -0,0 +1,26 @@
+---
+bc-version: [all]
```
JesperSchulz: The three-argument overload is available from runtime 4.2 (BC 15.2+), so `[all]` lets this rule and its bad sample be selected for older targets where the call does not compile. Please use `[16..]` for the first complete major that guarantees support, or `[15..]` only if this repository treats BC 15 as 15.2+.

community/knowledge/data-modeling/transferfields-skip-type-mismatch-can-drop-data.good.al
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+}
+
+codeunit 50491 "TransferFields Good"
+{
+    procedure CopyData(Source: Record "Transfer Source"; var Target: Record "Transfer Target")
+    begin
+        Target."Entry No." := Source."Entry No.";
+        Evaluate(Target."Reference", Source."Reference");
```
JesperSchulz: `Evaluate` is valid here and will fail loudly because its optional return value is omitted, but it performs conversion only; it does not invoke the destination field's `OnValidate` trigger. Since the article recommends handling conversion or validation, please either say this sample demonstrates conversion only or evaluate into a local `Integer` and call `Target.Validate("Reference", ConvertedReference)` when destination validation is required.

JesperSchulz (commented):
Two correctness gaps and one sample clarification.

community/knowledge/data-modeling/transferfields-skip-type-mismatch-can-drop-data.md
yahyatouil-dev: Fixed, reworded Best Practice to state that cross-extension type mismatches skip regardless of SkipFieldsNotMatchingType, and that any required field in that category needs explicit mapping. Also clarified the InitPrimaryKeyFields default. Thanks for catching this.

community/knowledge/data-modeling/transferfields-skip-type-mismatch-can-drop-data.good.al
yahyatouil-dev: Fixed, good.al now evaluates into a local variable and calls Target.Validate("Reference", ConvertedReference), so it demonstrates both conversion and validation instead of conversion only.

community/knowledge/data-modeling/transferfields-skip-type-mismatch-can-drop-data.md
yahyatouil-dev: Good catch, you're right, BC15.0 shipped with runtime 4.0 and 4.2 only arrives via a CU partway through the 15.x line, so [all] would've matched targets where this doesn't even compile. Bumped it to [16..], since BC16.0 ships with runtime 5.0.

yahyatouil-dev:
@JesperSchulz all three points addressed in the latest commits, ready for another look whenever you have time.

community/knowledge/data-modeling/transferfields-skip-type-mismatch-can-drop-data.md
JesperSchulz: Verified and resolved. The Best Practice section now states that `TransferFields(Source)` defaults `InitPrimaryKeyFields` to `true`, and that fields with no matching field number — plus fields whose types differ across extensions — are skipped regardless of `SkipFieldsNotMatchingType`, so anything the destination requires in those categories must be mapped and validated explicitly. Both points check out against the Learn reference for the overload.

community/knowledge/data-modeling/transferfields-skip-type-mismatch-can-drop-data.md
JesperSchulz: Resolved. `bc-version` is now `[16..]`, which is the right call — Learn lists the three-argument overload as "available or changed with runtime version 4.2", and since 4.2 only arrives partway through the 15.x line, BC 16.0 (runtime 5.0) is the first complete major that guarantees the call compiles.

community/knowledge/data-modeling/transferfields-skip-type-mismatch-can-drop-data.good.al
JesperSchulz: Resolved. `good.al` now evaluates into a local `ConvertedReference` and calls `Target.Validate("Reference", ConvertedReference)`, so the sample demonstrates conversion *and* destination validation. The omitted `Evaluate` return value still makes a bad value fail loudly rather than vanish.

JesperSchulz (approved):

PR #134, opened 2026-08-20
knowledge(performance): add community rules for performance

demiliani:
## Summary

Adds community performance guidance and paired AL examples for Business Central, covering partial-record behavior, transaction-safe HTTP deferral, multi-company access, page and session I/O, number-series batching, and database access patterns.

## Review updates

- use a transactional outbox for HTTP synchronization that must follow a committed write
- group cross-company customer lookups so `ChangeCompany` runs once per company
- guard UI-only FlowField styling from OData and Edit in Excel sessions
- remove the obsolete `CountApprox` guidance and the unsupported hidden FlowField rule
- refine samples to preserve behavior and compile against current AL APIs

## Validation

- `pwsh tools/Test-ReviewFixtures.ps1`
- `git diff --check`


community/knowledge/performance/countapprox-for-progress-not-count.md
```diff
@@ -0,0 +1,28 @@
...
+application-area: [all]
+---
+
+# Use CountApprox for progress UI, not Count
+
+> Contributions welcome — open a PR to refine or extend this article.
+
+## Description
+
+`Count()` asks SQL for an exact row count of the current filter. On a large table that is a `SELECT COUNT(*)` before any useful work starts — the usual cost of `Dialog.Open` with a percentage bar. `CountApprox()` exists for that UI case: it returns a cheap estimate (partition stats / metadata), accurate enough for a progress denominator. Agents default to `Count()` because the name matches "how many rows".
```
NKarolak: Are you sure about CountApprox? It is used in BCApps only 14 times, and not once for a progress bar.
If I remember correctly (but I am really not sure), it was said years ago that it does not provide that much value anymore. But it used to be important and useful in the past.

demiliani: Count() executes a precise database scan to return the exact number of records, whereas CountApprox() pulls from SQL Server statistics to return an estimated number significantly faster. In Progress Bar (UI) on large tables it’s faster. But yes, we’re talking about ms on rare cases now. The rule can also not be mandatory.

JesperSchulz:
I will get the eyes of some of our performance experts on this 🙂

community/knowledge/performance/oncompanyopen-subscribers-must-not-do-io.good.al
```diff
@@ -0,0 +1,17 @@
+codeunit 50100 "Login Subscriber IO Good"
+{
+    [EventSubscriber(ObjectType::Codeunit, Codeunit::"System Initialization", OnAfterLogin, '', false, false)]
+    local procedure OnAfterLogin()
+    begin
+        // Defer HTTP and heavy SQL to a job; session open must return immediately.
+        TaskScheduler.CreateTask(Codeunit::"Login Subscriber IO Work", Codeunit::"Login Subscriber IO Work", true, CompanyName(), CurrentDateTime());
```
JesperSchulz: This creates an unbounded task chain. A scheduled task starts a new background session and opens the company; `System Initialization`.OnAfterLogin is raised from `OnCompanyOpenCompleted` without a client-type guard. That session runs this subscriber and creates the next task. Please gate this to interactive client types and make scheduling idempotent, or move scheduling out of the login event. `CreateTask` also writes the Scheduled Task table during the session-open path that this rule says should avoid I/O. See [Task scheduler flow](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-task-scheduler#general-flow).

community/knowledge/performance/use-dedicated-lookup-pages-not-full-lists.bad.al
```diff
@@ -0,0 +1,43 @@
...
+            repeater(Rows)
+            {
+                field("No."; Rec."No.") { }
+                field(Name; Rec.Name) { }
+                field("Balance (LCY)"; Rec."Balance (LCY)") { }
+            }
+        }
+        area(factboxes)
+        {
+            part(Details; "Campaign Member List") { }
```
JesperSchulz: This is not a valid FactBox composition: the part references its own hosting page, whose `PageType` is `List`. Embedded parts must be separate `CardPart`/`ListPart` pages. Please use a distinct legal FactBox page so the bad sample still compiles. See [Page Parts Overview](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-designing-parts#different-types-of-parts).

community/knowledge/performance/validate-on-partial-record-forces-jit.good.al
```diff
@@ -0,0 +1,14 @@
+codeunit 50100 "Validate Partial Rec Good"
+{
+    procedure UppercaseUsCustomerNames()
+    var
+        Customer: Record Customer;
+    begin
+        Customer.SetRange("Country/Region Code", 'US');
+        if Customer.FindSet(true) then
+            repeat
+                Customer.Name := UpperCase(Customer.Name);
```
JesperSchulz: This changes behavior rather than only improving performance. `Customer.Name`'s `OnValidate` updates `Search Name` and calls `UpdateMyCustomer`; direct assignment bypasses both. Please use a field with no validation side effects for this example, or keep `Validate` and load the fields its trigger needs.

community/knowledge/performance/skip-setloadfields-on-write-and-transferfields.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Best Practice
+
+Use `SetLoadFields` only when the subsequent access is read-only. On a loop that writes the iterated record, or copies it with `TransferFields` / `Copy` onto a temporary record, omit `SetLoadFields` so the initial read already materializes every field those operations need.
+
+See sample: `skip-setloadfields-on-write-and-transferfields.good.al`.
+
+## Anti Pattern
+
+Calling `SetLoadFields` immediately before a `FindSet` whose body `Modify`s, `Delete`s, `Rename`s, or `TransferFields`s the same record. The review signal is partial-record setup on a record variable that is written or copied in the same iteration, not the mere presence of `SetLoadFields` on a read-only loop.
```
JesperSchulz: This signal classifies an existing Microsoft-layer good sample as bad: `microsoft/knowledge/performance/avoid-cloning-records-before-modify-delete-in-loops.good.al` uses `SetLoadFields` + `FindSet(true)` + `Validate` + `Modify(true)`. Also, the official partial-record guidance lists inserts, deletes, renames, field transfers, and temporary-record copies—not `Modify` itself—as operations requiring all fields. Please narrow this rule/sample to documented full-load operations or actual unloaded-field access so the corpus does not contradict itself. See [partial-record usage guidelines](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-partial-records#usage-guidelines).

community/knowledge/performance/countapprox-for-progress-not-count.good.al
```diff
@@ -0,0 +1,21 @@
...
+    var
+        Customer: Record Customer;
+        Window: Dialog;
+        Counter: Integer;
+        Total: Integer;
+    begin
+        Customer.SetRange("Country/Region Code", 'US');
+        Total := Customer.CountApprox();
+        Window.Open('Processing #1###### of #2######');
+        if Customer.FindSet(true) then
```
JesperSchulz: This loop never modifies `Customer`, so `FindSet(true)` unnecessarily reads the set with `UpdLock`. Please use `FindSet()` here; otherwise the good sample introduces a locking anti-pattern unrelated to the count optimization. See [`FindSet`'s `ForUpdate` semantics](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/methods-auto/record/record-findset-boolean-method).

community/knowledge/performance/countapprox-for-progress-not-count.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Best Practice
+
+Feed progress dialogs and informational messages with `CountApprox()`. Use `Count()` only when the exact integer is a business result (a posted control, a reconciliation, a test assertion).
+
+See sample: `countapprox-for-progress-not-count.good.al`.
+
+## Anti Pattern
+
+`Total := Rec.Count(); Window.Open(...);` immediately before a `FindSet` over the same filter. The exact count is discarded after the bar finishes; the user paid a full scan to draw it.
```
JesperSchulz: `Count()` does not necessarily require a full table scan: a filtered count can be served by SIFT when a SIFT key contains all filtered fields. Please qualify this claim to counts not supported by SIFT rather than teaching that every exact count scans the table. See [SQL methods and SIFT](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/administration/optimize-sql-al-database-methods-and-performance-on-server#calcfields-calcsums-and-count).

community/knowledge/performance/dataaccessintent-readonly-on-analytical-objects.md
```diff
@@ -0,0 +1,28 @@
...
+
+> Contributions welcome — open a PR to refine or extend this article.
+
+## Description
+
+Reports, API pages, and queries that only read can run against a read replica when `DataAccessIntent = ReadOnly`. Without the property they hit the primary replica and compete with posting. Agents omit it because the default is read-write and the object "only reads" in AL. The replica routing is a metadata switch, not something the compiler infers from the absence of `Modify`.
+
+## Best Practice
+
+On report, query, and API page objects that never write, set `DataAccessIntent = ReadOnly`. Keep the default on objects that insert, modify, or call a write codeunit from a processing-only report.
```
JesperSchulz: The applicability and version need narrowing. `DataAccessIntent` starts at runtime 5.0 (BC 16), so the frontmatter cannot be `[all]`. For pages it applies only to `PageType = API` with `Editable = false`; for queries it only affects OData/API exposure and has no effect in normal AL execution. Please use `[16..]` and state those constraints. See [DataAccessIntent property](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/properties/devenv-dataaccessintent-property).

JesperSchulz (commented):
High-confidence correctness findings only; community-layer placement is not a concern for this review.

community/knowledge/performance/avoid-currpage-update-in-onaftergetrecord.md
```diff
@@ -0,0 +1,28 @@
...
+
+> Contributions welcome — open a PR to refine or extend this article.
+
+## Description
+
+`OnAfterGetRecord` on a list already runs once per visible row on scroll and refresh. `CurrPage.Update` asks the page to reload, which fires those triggers again. The result is a refresh loop or a stutter on every row paint. Official developer performance guidance lists `CurrPage.Update()` in `OnAfterGetRecord` next to `Modify` as work that must not live there. Sibling of `do-not-modify-in-onaftergetrecord.md` (writes); this file is the client refresh half.
+
+## Best Practice
+
+Put display-only results in page variables assigned in `OnAfterGetRecord` without calling `Update`. If the page must refresh after an action, call `CurrPage.Update(false)` from `OnAction` or `OnAfterGetCurrRecord` once, not per row.
```
JesperSchulz: This conflicts with the Anti Pattern below: this line recommends `OnAfterGetCurrRecord`, while line 26 identifies that same location as a review signal when there is no explicit user action. `OnAfterGetCurrRecord` is not an explicit user action and can re-enter page refresh logic. Please restrict the recommendation to `OnAction` or another explicit action trigger.

community/knowledge/performance/prefer-related-table-over-extension-on-hot-ledgers.md
```diff
@@ -0,0 +1,28 @@
...
+application-area: [all]
+---
+
+# Prefer a related table over stored fields on hot ledgers
+
+> Contributions welcome — open a PR to refine or extend this article.
+
+## Description
+
+Stored fields on a table extension live in companion storage that is joined when the base row is read. On hot tables — G/L Entry, Item Ledger Entry, Cust. Ledger Entry — that join is paid on posting, lists, and APIs even when the extra columns are unused. A related table keyed by the ledger `Entry No.`, optionally surfaced with a FlowField or FactBox, leaves the base read path alone. Agents extend G/L Entry because it is "where the posting already is".
```
JesperSchulz: The claim that the join is paid even when extension fields are unused is incorrect for the named list and API paths. The platform automatically applies partial records to List/ListPart and OData pages, and explicitly excludes table extensions that have no fields selected for loading. Since v23, all extensions also share at most one companion-table join. Please scope this to paths that actually load extension fields or do not use partial records. See [partial-record usage guidelines](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-partial-records#usage-guidelines).

community/knowledge/performance/batch-number-series-instead-of-getnextno-per-row.md
```diff
@@ -0,0 +1,28 @@
...
+application-area: [all]
+---
+
+# Batch number-series calls instead of GetNextNo per insert
+
+> Contributions welcome — open a PR to refine or extend this article.
+
+## Description
+
+`Codeunit "No. Series".GetNextNo` updates and locks the number-series line on every call. A tight `Insert` loop that asks for a number per row serializes every concurrent writer on that series — the classic SaaS posting bottleneck. Training data still copies the per-row C/AL `NoSeriesManagement` shape. Codeunit `"No. Series - Batch"` issues numbers in memory and writes the series line once via `SaveState`. `NumberSequence` is the alternative when gaps are acceptable.
```
JesperSchulz: `GetNextNo` does not always update and lock the number-series line. The Sequence implementation used when gaps are allowed obtains numbers through `NumberSequence` and only modifies the line when the usage date or open state changes. Please scope the lock/contention claim and detection signal to the Normal/gapless implementation; as written it flags sequence-backed code that already avoids this contention.

community/knowledge/performance/httpclient-inside-write-transaction-holds-locks.md
```diff
@@ -0,0 +1,28 @@
...
+
+> Contributions welcome — open a PR to refine or extend this article.
+
+## Description
+
+The first database write opens an AL write transaction that the runtime holds until the execution completes or `Commit()` runs — see `understand-implicit-transaction-boundary.md`. `HttpClient` blocks the session until the remote call returns. Any locks taken by earlier `Insert`/`Modify`/`Delete` therefore stay held for the HTTP wall-clock time, and interactive users see a spinner. This is not generic "don't block": it is the AL transaction model plus lock lifetime around outbound I/O.
+
+## Best Practice
+
+Finish database writes and `Commit()` (or return from the write execution) before `HttpClient.Send`/`Get`/`Post`. If the call can be slow or retry, isolate it in a job queue or `TaskScheduler` task so UI and other sessions are not sitting on the writer's locks.
```
JesperSchulz: `Commit()` cannot be a generic remedy for the posting routines and subscribers mentioned below. It commits all prior work in the caller's transaction, so a later HTTP failure or outer-operation error can no longer roll that work back. Please make the good sample demonstrate deferral to a separate execution/task, or explicitly restrict `Commit()` to entry points where partial persistence is intentional. See [`Database.Commit`](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/methods-auto/database/database-commit-method).

community/knowledge/performance/countapprox-for-progress-not-count.bad.al
```diff
@@ -0,0 +1,22 @@
...
+        Customer: Record Customer;
+        Window: Dialog;
+        Counter: Integer;
+        Total: Integer;
+    begin
+        Customer.SetRange("Country/Region Code", 'US');
+        // Exact Count() is a SELECT COUNT(*) just to drive a progress bar.
+        Total := Customer.Count();
+        Window.Open('Processing #1###### of #2######');
+        if Customer.FindSet(true) then
```
JesperSchulz: Same issue as in the good sample: this loop only increments a counter and updates the dialog, so `FindSet(true)` takes an `UpdLock` on the whole filtered set for nothing. Please use `FindSet()` here as well — otherwise the fix is only half applied and the bad/good pair still shares an unrelated locking anti-pattern. See [`FindSet`'s `ForUpdate` semantics](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/methods-auto/record/record-findset-boolean-method).

community/knowledge/performance/validate-on-partial-record-forces-jit.good.al
```diff
@@ -0,0 +1,14 @@
+codeunit 50100 "Validate Partial Rec Good"
```
JesperSchulz, line 1: This sample is identical to `skip-setloadfields-on-write-and-transferfields.good.al` apart from the codeunit name and the comment, so two separate rules resolve to the same code. It also drops `SetLoadFields` entirely, which means it demonstrates the sibling rule (don't use partial records on a write path) rather than this one — the good sample never exercises a partial record at all, so it cannot show how to avoid the JIT.

Please make this sample keep the partial load and select the fields the code actually needs, so the contrast with `.bad.al` is the field selection rather than the absence of the feature. If the two rules genuinely collapse to the same advice, consider merging them instead.

community/knowledge/performance/use-dedicated-lookup-pages-not-full-lists.bad.al
demiliani: Fixed in updated PR.

community/knowledge/performance/validate-on-partial-record-forces-jit.good.al
demiliani: Fixed in updated PR.

community/knowledge/performance/skip-setloadfields-on-write-and-transferfields.md
demiliani: Fixed in updated PR.

community/knowledge/performance/countapprox-for-progress-not-count.good.al
demiliani: Fixed in updated PR.

community/knowledge/performance/oncompanyopen-subscribers-must-not-do-io.good.al
demiliani: Fixed in updated PR.

community/knowledge/performance/countapprox-for-progress-not-count.md
demiliani: Good catch. Fixed in next PR.

community/knowledge/performance/dataaccessintent-readonly-on-analytical-objects.md
demiliani: Fixed in updated PR.

community/knowledge/performance/avoid-currpage-update-in-onaftergetrecord.md
demiliani: Fixed in an updated PR.

community/knowledge/performance/prefer-related-table-over-extension-on-hot-ledgers.md
demiliani: Fixed on updated PR.

community/knowledge/performance/batch-number-series-instead-of-getnextno-per-row.md
demiliani: Rule improved on next PR.

community/knowledge/performance/httpclient-inside-write-transaction-holds-locks.md
demiliani: Fixed in updated PR (correct deferral pattern: the write stays in the caller's transaction (locks released when it ends naturally), and a TaskScheduler task runs the HTTP call in a separate session where no write-transaction lock is held.)

community/knowledge/performance/countapprox-for-progress-not-count.bad.al
demiliani: Fixed.

community/knowledge/performance/validate-on-partial-record-forces-jit.good.al
demiliani: Fixed.

community/knowledge/performance/avoid-currpage-update-in-onaftergetrecord.md
JesperSchulz: Verified in 5f93286: `OnAfterGetCurrRecord` was removed from the recommendation. Resolved.

community/knowledge/performance/countapprox-for-progress-not-count.good.al
JesperSchulz: Verified in 5f93286: the read-only loop now uses `FindSet()`. Resolved.

community/knowledge/performance/countapprox-for-progress-not-count.md
JesperSchulz: Verified in 5f93286: the article now accounts for SIFT-covered counts. Resolved.

community/knowledge/performance/prefer-related-table-over-extension-on-hot-ledgers.md
JesperSchulz: Verified in 5f93286: the article now accounts for partial-record exclusion and the v23 companion-table model. Resolved.

community/knowledge/performance/batch-number-series-instead-of-getnextno-per-row.md
JesperSchulz: Verified in 5f93286: the contention claim and detection signal are now limited to gapless/Normal series. Resolved.

community/knowledge/performance/validate-on-partial-record-forces-jit.good.al
JesperSchulz: Verified in 5f93286: the good sample now keeps the partial load and includes the fields needed by `Name.OnValidate`. Resolved.

community/knowledge/performance/validate-on-partial-record-forces-jit.good.al
JesperSchulz: Verified in 5f93286: the sample retains `Validate` and loads `Search Name`, preserving the field trigger behavior. Resolved.

community/knowledge/performance/countapprox-for-progress-not-count.bad.al
JesperSchulz: Verified in 5f93286: the bad sample's read-only loop now uses `FindSet()`. Resolved.

community/knowledge/performance/skip-setloadfields-on-write-and-transferfields.md
JesperSchulz: Verified in 5f93286: `Modify` was removed from the blanket rule and the sample now uses documented `TransferFields`/temporary-copy behavior. Resolved.

community/knowledge/performance/use-dedicated-lookup-pages-not-full-lists.bad.al
JesperSchulz: Verified in 5f93286: the FactBox now references a separate `CardPart`. Resolved.

community/knowledge/performance/dataaccessintent-readonly-on-analytical-objects.md
JesperSchulz: Verified in 5f93286: version scope and API page/query applicability are now accurate. Resolved.

community/knowledge/performance/oncompanyopen-subscribers-must-not-do-io.good.al
JesperSchulz: Still open in `5f93286`: the revised sample does not compile. The platform method is `Session.CurrentClientType()`, not `Session.GetCurrentClientType()`. Also, `TaskScheduler.TaskExists` accepts the task GUID returned by `CreateTask`; `Codeunit::"Login Subscriber IO Work"` is an integer object ID and cannot implement this check. Please use the correct client-type method and persist/query a real task identifier for idempotency.

community/knowledge/performance/httpclient-inside-write-transaction-holds-locks.md
JesperSchulz: The transaction boundary is fixed, but the revised sample changes the operation's scope. `SyncCustomerLastName` receives one customer, while the scheduled task receives no `RecordId` and runs `FindSet()` over every customer; every call therefore schedules a full-customer sync. Please pass `Customer.RecordId` to `CreateTask` and bind the task codeunit to `Customer` (or persist an explicit work item) so deferral preserves the original one-customer behavior.

demiliani:
Updated PR submitted with all the remaining suggested fixes.

community/knowledge/performance/httpclient-inside-write-transaction-holds-locks.md
JesperSchulz: Verified in 8373bc4: the task now receives `Customer.RecordId`, declares `TableNo = Customer`, and processes only that bound record. Resolved.

community/knowledge/performance/oncompanyopen-subscribers-must-not-do-io.good.al
JesperSchulz: Verified in 8373bc4: the sample now uses `Session.CurrentClientType()`, stores the GUID returned by `CreateTask`, and passes that GUID to `TaskExists`. Resolved.

JesperSchulz (approved):
Nice work! I think a few of these should get promoted to the Microsoft layer, but let's get them in first. I'll take a look all up at the community layer once a month I think, and pull the strongest articles one layer up 🎉

community/knowledge/performance/countapprox-for-progress-not-count.md
BardurKnudsen: Natalie is correct. The intention was that CountApprox() should be faster than Count(). And it often was. Unfortunately the 'approx'-part meant that it sometimes returned 0 even if there were records. That meant two things: 1) We couldn't use the result for deciding whether to process anything or not, and 2) when we want to show progress as %, it meant that we potentially risked a division by zero, which again meant that we had to add extra code lines to guard against it.
Anyways: I just checked with our platform guys, and they tell me that CountApprox() calls the same code as Count() and has done so for a decade or so.

community/knowledge/performance/changecompany-in-loop-drops-caches.good.al
```diff
@@ -0,0 +1,25 @@
...
+    var
+        Customer: Record Customer;
+    begin
+        Customer.ChangeCompany(CompanyNameValue);
+        Customer.SetLoadFields(Name);
+        if Customer.Get(CustomerNo) then
+            exit(Customer.Name);
+    end;
+
+    procedure NamesForCompanies(var Company: Record Company)
```
BardurKnudsen: Not sure why this pattern is better regarding change company - other than the SetLoadFields, which perhaps could have been set outside the loop. Is the function NameInCompany used?

community/knowledge/performance/countapprox-for-progress-not-count.good.al
```diff
@@ -0,0 +1,21 @@
...
+{
+    procedure RecalcUsCustomers()
+    var
+        Customer: Record Customer;
+        Window: Dialog;
+        Counter: Integer;
+        Total: Integer;
+    begin
+        Customer.SetRange("Country/Region Code", 'US');
+        Total := Customer.CountApprox();
```
BardurKnudsen: Same as count()

community/knowledge/performance/guiallowed-guard-on-pages-used-as-odata.good.al
```diff
@@ -0,0 +1,25 @@
...
+                field(Name; Rec.Name) { }
+            }
+        }
+    }
+
+    trigger OnAfterGetRecord()
+    begin
+        if not GuiAllowed then
+            exit;
+        Rec.CalcFields("Balance (LCY)");
```
BardurKnudsen, line 29: Would this ever be good when we don't expose the balance field anyway?

community/knowledge/performance/httpclient-inside-write-transaction-holds-locks.good.al
```diff
@@ -0,0 +1,25 @@
+codeunit 50100 "HttpClient Holds Locks Good"
+{
+    // Write completes inside the caller's transaction; HTTP deferred so locks are released with it.
+    procedure SyncCustomerLastName(var Customer: Record Customer)
+    begin
+        Customer."Search Name" := Customer.Name;
+        Customer.Modify(false);
+        // RecordId binds the task to this specific customer; the platform loads it into Rec on OnRun.
+        TaskScheduler.CreateTask(Codeunit::"Customer Sync Task", 0, true, CompanyName(), CurrentDateTime(), Customer.RecordId);
```
BardurKnudsen: Generally, we should not lock while we do lengthy stuff like calling webservices etc.
In this concrete example you may risk starting a sync to before the data is committed, so maybe you sync the new (uncommitted) value or maybe you sync the old name. And if the transaction rolls back due to some error, you already made sync call. So perhaps syncs to external systems should happen after data is committed?

community/knowledge/performance/query-results-bypass-primary-key-cache.bad.al
```diff
@@ -0,0 +1,28 @@
...
+}
+
+codeunit 50100 "Query Bypass PK Cache Bad"
+{
+    procedure CustomerName(CustomerNo: Code[20]): Text
+    var
+        CustomerByNo: Query "Query Bypass PK Cache Bad Q";
+    begin
+        // Query Open/Read never hits the server PK cache.
+        CustomerByNo.SetRange(NoFilter, CustomerNo);
```
BardurKnudsen, line 22: True. But an even simpler example that also omits the PKC is Customer.FindFirst()

community/knowledge/performance/visible-false-does-not-skip-page-field-cost.md
```diff
@@ -0,0 +1,28 @@
...
+application-area: [all]
+---
+
+# Visible false does not skip page-field cost
+
+> Contributions welcome — open a PR to refine or extend this article.
+
+## Description
+
+`Visible = false` and `Enabled = false` hide a control; they do not remove it from the page metadata the client and server still process. List pages in particular still load bound fields and can still calculate FlowFields on those controls — see `hidden-flowfields-still-calculate-before-bc26-opt-in.md` for the FlowField-specific opt-in. Official page-performance guidance is to **delete** the field from the page object when users do not need it. Agents hide heavy columns instead of removing them.
```
BardurKnudsen: Well; in newer versions of BC (and also in old C/SIDE client, if anyone still uses it) only visible flowfields are calculated. By removing the expense field from the page you also prevent the user from showing it.

BardurKnudsen (approved):

NKarolak (changes requested):

community/knowledge/performance/countapprox-for-progress-not-count.md
demiliani: I will remove this rule if unuseful then.

community/knowledge/performance/visible-false-does-not-skip-page-field-cost.md
demiliani: This rule also can be removed for newer BC versions.

community/knowledge/performance/httpclient-inside-write-transaction-holds-locks.good.al
demiliani: I updated the example to use a transactional outbox in httpclient-inside-write-transaction-holds-locks.good.al:
- Customer update and outbox entry commit or roll back together.
- A recurring job queue processes only committed entries.
- HTTP runs before the worker deletes the entry, avoiding write locks.
- The guidance now notes retry/idempotency requirements.

community/knowledge/performance/changecompany-in-loop-drops-caches.good.al
demiliani: Improved the sample with comparable good and bad.

community/knowledge/performance/guiallowed-guard-on-pages-used-as-odata.good.al
demiliani: Good point. In the previous sample, the calculated balance was unused, so removing the calculation would have been better than guarding it. I updated both samples so balance controls a UI-only style expression. The GuiAllowed guard now prevents that presentation-only calculation from running for OData and Edit in Excel.

NKarolak (approved):

PR #137, opened 2026-08-24
Add community guidance and review support for Business Central agents

demiliani:
## Summary
Adds a new community knowledge domain for reviewing Business Central Agent SDK implementations.

The contribution includes 20 guidance rules, each with documentation and good/bad AL examples, covering:

- Agent registration and interface wiring
- Profiles and setup-page design
- Setup persistence and user identity
- Least-privilege permissions and effective access
- Instruction structure, keywords, and secure storage
- Agent lifecycle and upgrade behavior
- Session-bound subscribers and input review
- Cross-app APIs and Copilot capability registration
- Programmatic agent creation safeguards
- Error and warning handling

- Adds the standalone community-owned `al-agents-review` action skill, dispatched by Entry alongside `al-code-review`.
- Keeps the Agents domain under the `community` layer.


demiliani:
Anyone can check what gate is blocking this PR? Thanks in advance.

JesperSchulz:
> Anyone can check what gate is blocking this PR? Thanks in advance.

All green. We just need some reviews ☺️ Let me get on that, and ask others to join the fun!

community/skills/review/al-agents-review.md
```diff
@@ -0,0 +1,68 @@
+---
+kind: action-skill
+id: al-agents-review
+version: 1
+title: AL agents review
+description: Reviews AL source changes against agent guidance from BCQuality.
```
JesperSchulz, line 6: This lands as a top-level peer of `al-code-review`, not as one of its sub-skills — and I think that needs saying explicitly here.

Tracing it through: `entry.md:53` sources all action skills under `*/skills/**/*.md` across enabled layers, so this file is discovered. At `entry.md:71` the 16 microsoft leaves get dropped as `superseded-by-super-skill` because they're listed in `al-code-review`'s `sub-skills`. This one isn't listed, so it survives and gets dispatched alongside the super-skill. `entry.md:72` layer precedence doesn't fire either, since that only applies to candidates sharing an `id`.

That's fine as a mechanism, and given the `microsoft` layer can't reference `community` it's the only one available. But it has consequences worth documenting:

- Two independent findings-reports instead of one rollup, so no cross-report dedup and no `<from-sub-skill>:` id prefixing.
- `al-code-review`'s self-review pass validates its own agent findings against BCQuality knowledge, but it never sources `community/knowledge/agents/`. An agent issue it spots on its own can't be cited or suppressed against these new articles, so you can end up with an untethered `agent:` finding sitting next to a duplicate knowledge-backed one from this leaf.
- Dispatch depends on goal-match scoring rather than an explicit list, which is a softer guarantee than the microsoft leaves get.

Every microsoft leaf carries the sentence "It is one of the skills composed by `al-code-review`." Could you add the inverse here, so the next reader doesn't assume composition?

tools/Test-ReviewFixtures.ps1
```diff
@@ -143,27 +201,20 @@ foreach ($domain in $leafDomains) {
...
+        $selectedArticle = $articles | Select-Object -First 1
     }
     if (-not $selectedArticle) {
         $problems.Add("${domain}: no article has both .good.al and .bad.al companion samples.") | Out-Null
         continue
     }
 
-    $articlePath = "microsoft/knowledge/$domain/$($selectedArticle.Name)"
+    $articlePath = [string]$selectedArticle.ArticlePath
+    $sampleDirectory = Split-Path -Parent $articlePath
```
JesperSchulz: This breaks the harness on Windows.

`$articlePath` is forward-slashed, but `Split-Path -Parent` returns backslashes, so `$sampleDirectory` comes back mixed-separator. Plain validation still passes because `Test-Path` tolerates it — but `-PrepareDirectory`, which is what CI actually runs (`.github/workflows/review-fixtures.yml:18`), throws at line 405:

```
agents-bad: deterministic ranking omitted selected article
'community\knowledge\agents/agent-permissions-intersect-with-assigner.md'.
```

Linux CI is unaffected, which is why the checks on this PR are green.

I verified the fix locally — with the normalization added, the full prepare run passes: `Prepared 34 neutral fixture(s)`, `PASSED: 34 cases cover 17 leaf domains`. Everywhere else in this file already normalizes after `GetRelativePath`; this one spot got missed.

```suggestion
    $sampleDirectory = (Split-Path -Parent $articlePath).Replace('\', '/')
```

tools/Test-ReviewFixtures.ps1
```diff
@@ -143,27 +201,20 @@ foreach ($domain in $leafDomains) {
         if ($articleName.EndsWith('.md')) {
             $articleName = [System.IO.Path]::GetFileNameWithoutExtension($articleName)
         }
-        $candidate = Join-Path $knowledgeDirectory "$articleName.md"
-        if (Test-Path -LiteralPath $candidate -PathType Leaf) {
-            $selectedArticle = Get-Item -LiteralPath $candidate
-        } else {
+        $selectedArticle = $articles | Where-Object BaseName -eq $articleName | Select-Object -First 1
```
JesperSchulz, line 204: Small side-effect of the refactor: this now tests membership in the paired-samples list rather than `Test-Path` on the `.md`.

So an override naming a real article that happens to lack `.good.al`/`.bad.al` companions reports `override article does not exist` — which isn't true — and then falls through to line 212 and reports `no article has both .good.al and .bad.al companion samples` as well. Two errors for one problem, and the first one sends you looking for a typo that isn't there.

community/knowledge/agents/bind-agent-subscribers-only-in-agent-session.bad.al
```diff
@@ -0,0 +1,9 @@
+codeunit 50101 "Sales Review Agent Events"
```
JesperSchulz, line 1: This codeunit needs `EventSubscriberInstance = Manual;`.

The paired `.good.al` calls `BindSubscription(GlobalAgentEvents)` against this type, and with the default `StaticAutomatic` that's an error rather than a no-op:

> With this setting, you can **not** call the `BindSubscription` method or `UnBindSubscription` method for any events in this codeunit; otherwise, an error occurs.
> — `devenv-eventsubscriberinstance-property`

Both real agent examples declare it: `SOA - KPI Track All` (`SOAKPITrackAll.Codeunit.al`) and `SOA Shipment Date Mgt.` (`SOAShipmentDateMgt.Codeunit.al`) in ALAppExtensions.

Related: the good sample references a codeunit that only exists in this bad sample, so it can't be read on its own. It might be worth giving the good sample its own bound codeunit with `Manual` set — that also makes the property the visible difference between the pair, which is really the point of the article.

community/knowledge/agents/do-not-create-agents-in-install-upgrade-or-background.good.al
```diff
@@ -0,0 +1,46 @@
...
+    trigger OnQueryClosePage(CloseAction: Action): Boolean
+    var
+        Agent: Codeunit Agent;
+        TempAgentAccessControl: Record "Agent Access Control" temporary;
+        AgentUserSecurityId: Guid;
+    begin
+        if CloseAction = CloseAction::Cancel then
+            exit(true);
+
+        AgentUserSecurityId := Agent.Create(
```
JesperSchulz: Calling `Agent.Create` directly from `OnQueryClosePage` isn't the canonical setup-page flow. Creation goes through `AgentSetup.SaveChanges(buffer)`, which creates-or-updates and returns the User Security ID.

BCTech `SalesValAgentSetup.Page.al`:

```al
CurrPage.AgentSetupPart.Page.GetAgentSetupBuffer(TempAgentSetupBuffer);
Rec."User Security ID" := SalesValAgentSetup.SaveAgent(TempAgentSetupBuffer);
```

which routes to `AgentSetup.SaveChanges(TempAgentSetupBuffer)`. ALAppExtensions `SOASetup.Page.al` does the same, and the `ai-agent-sdk-setup-page` doc shows the same shape. Going straight to `Agent.Create` here bypasses the buffer's change tracking and the `Agent Setup Part` integration — `Agent.Create` is the lower-level primitive for automation scenarios.

It also contradicts `agent-setup-source-table-is-temporary.good.al`, which declares the same `page 50100 "Sales Review Agent Setup"` object name and ID but correctly goes through `GetChangesMade`/`SaveChanges`. Two good samples, same object, opposite advice.

Separately: `TempAgentAccessControl` is declared and passed but never populated, so the agent gets created with nobody holding management access over it.

community/knowledge/agents/analyze-message-error-stops-warning-forces-review.good.al
```diff
@@ -0,0 +1,26 @@
+codeunit 50100 "Sales Review Agent Task"
+{
+    procedure AnalyzeAgentTaskMessage(AgentTaskMessage: Record "Agent Task Message"; var Annotations: Record "Agent Annotation")
+    var
+        AgentMessage: Codeunit "Agent Message";
+        NotRelevantMsg: Label 'Message is not a sales order task.';
+        NotRelevantDetailsTxt: Label 'Provide a message related to sales order review.';
+    begin
+        if AgentTaskMessage.Type = AgentTaskMessage.Type::Output then begin
+            AgentMessage.UpdateText(AgentTaskMessage, AgentMessage.GetText(AgentTaskMessage) + '\n\nWritten with the help of AI');
```
JesperSchulz: `'\n\n'` won't produce line breaks. AL has no C-style escapes, so this appends literal backslash-n characters into the outbound message text. For real newlines in data use `#13#10` (or `TypeHelper.GetNewLine()`); `\` only means a paragraph break inside `Message`/`Error`/`Confirm` dialog text, which isn't what's happening here.

In fairness, this is copied straight from the `ai-agent-sdk-define-register` doc, which has the same bug — worth reporting upstream. Microsoft's actual SOA code doesn't concatenate a disclaimer at all; it merges the signature into HTML content.

Two other things while we're in this sample:

- The article's headline rule is "Error stops the task; Warning still requires review", but only `Severity::Warning` is ever set. The Error branch never appears, so the sample doesn't demonstrate the distinction the article is teaching.
- `SOAAnnotation.Codeunit.al` calls `Clear(Annotations)` before populating each annotation. Since `Annotations` arrives here as a `var` parameter, skipping that risks carrying stale field values into the `Insert()`.

community/knowledge/agents/use-documented-instruction-keywords.md
```diff
@@ -0,0 +1,30 @@
...
+technologies: [al]
+countries: [w1]
+application-area: [all]
+---
+
+# Use the toolkit instruction keywords for review, mail, and memory
+
+## Description
+
+The agent runtime looks for specific phrases: ask for assistance, request a review, reply, write an email, memorize, set a field, use lookup, invoke an action. Ordinary English such as get a human to look or remember this is weaker. Outbound reply and email always require review; that is platform policy, not optional tone.
```
JesperSchulz: Two of the eight are paraphrased rather than quoted — the documented keywords are `Set field` and `Invoke action`, but this reads "set a field" and "invoke an action".

In an article whose entire point is to use the documented keywords verbatim, that's a little self-defeating.

community/knowledge/agents/agent-setup-page-is-configuration-dialog.md
```diff
@@ -0,0 +1,27 @@
...
+# Agent setup pages use ConfigurationDialog and the Agent Setup Part
+
+## Description
+
+Instance setup is not a Card or StandardDialog. The toolkit expects `PageType = ConfigurationDialog` so OK and Cancel are system actions, plus the built-in `Agent Setup Part` for name, display name, state, and access. A Card with custom fields only drops those shared controls and the AI-use notices the part carries.
+
+## Best Practice
+
+Declare `PageType = ConfigurationDialog`, host `part(...; "Agent Setup Part")`, and put agent-specific fields in another group. Keep system OK/Cancel. Use a temporary source record and defer persistence until Update, as described in `agent-setup-source-table-is-temporary.md`.
+The Extensible property of the page must be set to false.
```
JesperSchulz: "must" is stronger than the docs support — I couldn't find anywhere that mandates `Extensible = false` on the setup page. It's a near-universal convention in Microsoft's samples, so "should", or "Microsoft's samples consistently set...", would land better.

Also, this sentence arrives after the Best Practice paragraph and splits it from the `See sample:` line. Reads more naturally folded into the paragraph above.

community/knowledge/agents/register-copilot-capability-for-the-agent.good.al
```diff
@@ -0,0 +1,26 @@
...
+    trigger OnInstallAppPerDatabase()
+    var
+        CopilotCapability: Codeunit "Copilot Capability";
+        LearnMoreUrlTxt: Label 'https://example.com/sales-review-agent', Locked = true;
+    begin
+        if not CopilotCapability.IsCapabilityRegistered(Enum::"Copilot Capability"::"Sales Review Agent") then
+            CopilotCapability.RegisterCapability(
+                Enum::"Copilot Capability"::"Sales Review Agent",
+                Enum::"Copilot Availability"::Preview,
+                "Copilot Billing Type"::"Microsoft Billed",
```
JesperSchulz: Nit: `Enum::"Copilot Availability"::Preview` on the line above is prefixed but this one isn't. Worth matching them.

community/knowledge/agents/wire-all-three-agent-interfaces.good.al
```diff
@@ -0,0 +1,10 @@
+enumextension 50100 "Sales Review Agent Metadata" extends "Agent Metadata Provider"
+{
+    value(50100; "Sales Review Agent")
+    {
+        Caption = 'Sales Review Agent';
+        Implementation = IAgentFactory = "Sales Review Agent Factory",
+                         IAgentMetadata = "Sales Review Agent Metadata",
```
JesperSchulz: Nit: the enumextension on line 1 and this `IAgentMetadata` implementation codeunit share the name "Sales Review Agent Metadata". Legal, since object names are unique per object type, but `IAgentMetadata = "Sales Review Agent Metadata"` reads ambiguously right underneath `enumextension 50100 "Sales Review Agent Metadata"`. Something like "Sales Review Agent Meta. Impl." for the codeunit would disambiguate.

JesperSchulz (changes requested):
Thorough pass over this one — read all 20 articles and 40 samples, ran every CI validator locally on Windows, and verified the Agent SDK claims against MS Learn, `microsoft/ALAppExtensions` (`f86dcaf`) and `microsoft/BCTech` (`7b1360e`).

**The knowledge itself is good.** It's genuinely remedial and BC-specific, there's zero overlap with `microsoft/knowledge`, every article carries a `Detection signal:` line, and the `## See also` cross-links are used well. 16 of the 17 factual claims I checked came back verified — including the subtle one I expected to be wrong: `Agent.Create` takes `Record "Agent Access Control"` while `GetDefaultAccessControls` takes `Record "Access Control Buffer"`. Those really are two different records for two different concerns, and all six samples get it right.

Requesting changes on two blocking items plus a few sample-correctness fixes. Details inline; three things that have no line to attach to are below.

---

### The PR description overstates what's wired up

"Enables Agents findings in AL code reviews" implies this leaf runs inside `al-code-review`. It doesn't — `al-code-review.md:42` says sub-skills are never discovered implicitly, and the `microsoft` layer can't reference `community`, so it never will.

What actually happens is that Entry dispatches it as a **top-level peer** of `al-code-review`. That works, but it's a different mechanism with different properties (see the comment on `al-agents-review.md`). Could you reword to something like *"Adds a standalone `al-agents-review` action skill, dispatched by Entry alongside `al-code-review`"*?

Not a blocker for this PR, but worth noting: there's currently no sanctioned way for **any** community review leaf to join a microsoft super-skill's rollup. Cross-layer composition only works downward, and `custom` lives in consumer repos. If community review leaves are meant to be a supported pattern, `al-code-review` probably needs an extension point — sourcing sub-skills by domain across enabled layers rather than by hardcoded path. Happy to open a separate issue.

### The auto-selected evaluation fixture is the weakest pair in the set

`evaluation/review-fixtures.json` has no `overrides` entry for `agents`, so the harness takes the first article alphabetically: `agent-permissions-intersect-with-assigner`. Its `.bad.al` is a bare 12-line permissionset with no agent context at all — more naturally an `al-security-review` finding — and the `.good.al` is a 7-line permissionset.

With `minimumExpectedRecall: 1.0` and `minimumCleanRate: 1.0` that's a fragile control. 10 of the 16 existing domains already carry an override for exactly this reason. `wire-all-three-agent-interfaces` or `do-not-create-agents-in-install-upgrade-or-background` would both be much sharper.

### Smaller things

- `README.md:93` — the Scope paragraph enumerates the covered domains and wasn't updated for agents.
- `bc-version` — 18 articles are `[27..]`, but the SDK is sandbox-only from 27.4 and production-supported from 28.1, so `[27..]` over-claims 27.0–27.3. The schema only supports majors, so this may just be unavoidable; flagging it in case you want the two `[28..]` articles to set the floor for everything.
- `al-agents-review.md` doesn't define a `not-applicable` gate even though it lists that outcome. `al-ui-review.md` has an explicit file-type gate; without one, this fairly niche leaf evaluates every AL diff. (The missing `json` Output example is fine — `al-query-review.md` has none either.)

### CI note

`flag` and `guard` show as failing in the checks list, but both are actually `cancelled` — they're `pull_request_target` workflows from a fork waiting on maintainer approval, not content failures. `validate`, `validate-index` and `validate-review-fixtures` all genuinely pass.

JesperSchulz (approved):
Re-reviewed at `e98fb8e`. Everything from the previous pass is addressed — verified each fix against the diff, and re-ran the full validator suite on Windows including the `-PrepareDirectory` path that was broken before.

**Blocking items — both resolved**

- `Test-ReviewFixtures.ps1:230` now normalizes the separator. I re-ran the CI mode on Windows and it passes clean: `Prepared 34 neutral fixture(s)` / `PASSED: 34 cases cover 17 leaf domains`.
- `al-agents-review.md` now states the dispatch relationship outright ("It is not one of the skills composed by `al-code-review`; Entry discovers and dispatches it as a top-level peer"), and the PR description matches. Thanks for taking the wording.

**Sample correctness — all three fixed correctly**

- The bind pair is now coherent: the good sample defines its own `codeunit 50101 "Sales Review Agent Subscribers"` with `EventSubscriberInstance = Manual`, and the bad sample is left as a plain static subscriber. The property is now the visible difference between the two, which is exactly what the article is teaching.
- `do-not-create-agents…good.al` goes through `GetAgentSetupBuffer` → `AgentSetup.SaveChanges`, matching BCTech `SalesValAgentSetup.Page.al`. It no longer contradicts `agent-setup-source-table-is-temporary.good.al`, and the unpopulated access-control record is gone.
- `analyze-message…good.al` uses `#13#10`, demonstrates both the Error and Warning branches, and adds `Clear(Annotations)` before each populate. `IsRelevant` doing a real substring test instead of `<> ''` is a nice extra — the clean control is more convincing for it.

**The fixture override is a big improvement**

Swapping in `wire-all-three-agent-interfaces` gives the harness a genuinely sharp pair. The generated positive control is an enumextension wiring only `IAgentFactory`, against a clean control wiring all three — unambiguous in both directions, which is what you want at `minimumExpectedRecall: 1.0` / `minimumCleanRate: 1.0`.

Also nice: the `not-applicable` gate I mentioned as a nit got a real definition rather than a hand-wave, and the override error branch in `Test-ReviewFixtures.ps1` now distinguishes "doesn't exist" from "exists but has no companions" and `continue`s, so it no longer double-reports.

All six checks are green, and `flag`/`guard` now run for real rather than sitting cancelled.

One cosmetic leftover, definitely not worth another round: `register-copilot-capability-for-the-agent.bad.al:7` still has `IAgentMetadata = "Sales Review Agent Metadata"` from before the rename. Harmless in a bad sample — fold it in if you touch the file for something else.

Nice contribution. LGTM.

PR #139, opened 2026-08-27
knowledge(breaking-changes): adding a parameter to an event publisher is not a signature break

gggdttt:
`breaking-changes/do-not-change-published-procedure-signatures.md` opens its scope with "Signature changes include **adding**, removing, or reordering parameters" and its Detection reads "a parameter **added**, removed, reordered, retyped, or flipped to/from `var` ... on any non-`local` procedure that already shipped". Applied to an event publisher, that reports a plain parameter addition as a breaking change.

That is the source of a false positive on BCApps PR 10278, where adding a parameter to an existing `[IntegrationEvent]` was reported as breaking existing subscribers.

## Why it is wrong

AL binds an event subscriber to its publisher **by parameter name, not by position**. An added parameter is simply not declared by existing subscribers, and they keep binding. `events/adding-a-parameter-to-an-event-is-not-a-breaking-change` already says this; the breaking-changes article contradicted it, and a knowledge file that actively asserts the opposite defeats the suppression the events article is supposed to provide.

Verified in shipping BCApps code rather than from documentation — `Test Runner - Mgt` publishes `OnBeforeTestMethodRun(var CurrentTestMethodLine; CodeunitID; CodeunitName; FunctionName; FunctionTestPermissions; var Skip)` and `ALTestRunnerResetEnvironment` binds to it declaring `(CodeunitID; CodeunitName; FunctionName; FunctionTestPermissions; var CurrentTestMethodLine)`, omitting one parameter and reordering another.

## Changes

- **Description** — a new paragraph separates procedures dependents *call* from event publishers they *bind to*, states the by-name binding rule, and routes the reader to `events/adding-a-parameter-to-an-event-is-not-a-breaking-change` for the full treatment and to `events/add-new-event-parameters-at-the-end` for the case where publisher access does make an addition breaking. Adding `var IsHandled: Boolean` is routed to `events/do-not-add-ishandled-to-an-existing-event`, since it binds fine but changes the event contract.
- **Anti Pattern** — Detection now excludes an event publisher whose only change is an added parameter, and names that report as a false positive.
- `integration-event` added to `keywords`.

Everything the article says about ordinary non-`local` procedures is unchanged, and removals, retypes, renames, and `var` flips on event publishers remain in scope.

## Note on the revision history

The first commit on this branch limited the carve-out to parameters *appended at the end* and justified it with leading-prefix binding. That was wrong on both counts and was corrected in `07a2b34`. Only the final state should be reviewed.

## Related

#138 removes the mirror-image contradiction from `style/event-subscriber-param-names-match-publisher`.


microsoft/knowledge/breaking-changes/do-not-change-published-procedure-signatures.md
```diff
@@ -13,6 +13,8 @@ application-area: [all]
...
 
+This rule governs procedures that dependents *call*. An event publisher — a procedure carrying `[IntegrationEvent]` or `[BusinessEvent]`, conventionally declared `local` — is bound to, not called, and binds on a leading prefix of its parameter list. Appending a new parameter at the end of a shipped event therefore leaves every existing subscriber binding successfully, so it is additive rather than breaking and must not be flagged under this rule. See `events/add-new-event-parameters-at-the-end`. Every other edit to a published event signature — removing, reordering, or retyping a parameter, or flipping one to or from `var` — still breaks binding and is in scope here. Appending `var IsHandled: Boolean` is a separate concern: it binds fine but changes the event's contract, and is covered by `events/do-not-add-ishandled-to-an-existing-event`.
```
pri-kise: you can also add a parameter in between if the event procedure is local (not accessible for other apps).
The only importance about events is the name of the published paramters of the procedure and the procedure name

The var part is correct and is already covered by
https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/analyzers/appsourcecop-as0063
and
https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/analyzers/appsourcecop-as0077
(not sure if this should be mentioned here)

gggdttt:
Thanks, both points are now handled.

**Parameter position.** Agreed, and the article no longer claims appending is required. It now states that AL binds each subscriber parameter by name rather than by position, so a parameter added to a shipped event leaves existing subscribers binding successfully *at any position in the list*. I verified this against shipping BCApps code rather than documentation: `TestRunnerMgt.Codeunit.al` publishes `OnBeforeTestMethodRun(var CurrentTestMethodLine; CodeunitID; CodeunitName; FunctionName; FunctionTestPermissions; var Skip)` and `ALTestRunnerResetEnvironment.Codeunit.al` subscribes with `BeforeTestMethod(CodeunitID; CodeunitName; FunctionName; FunctionTestPermissions; var CurrentTestMethodLine)` — it omits `Skip` and moves the first parameter to the end. Scanning the whole repo, 121 comparable publisher/subscriber pairs break down as 110 leading-prefix, 11 reordered, and 0 declaring a parameter the publisher does not have.

The public-event caveat you are implying is kept: a public event procedure is also callable by dependents, so a required parameter added anywhere does break callers under AS0024. This article routes that case to `events/add-new-event-parameters-at-the-end`, which covers it.

**AS0063 / AS0077.** Good catch, and they belong in the knowledge base — but not duplicated here. `events/treat-local-and-internal-events-as-subscriber-contracts` already owns the `var` and rename cases and already names AS0025, AS0063 and AS0077. I have pushed a follow-up commit that stops this breaking-changes article from claiming those cases and routes them to that article instead, citing the three rule IDs so the link is discoverable from here.

JesperSchulz (approved):

PR #142, opened 2026-08-27
Add AL-focused AppSource validation guidance

demiliani:
## Summary

- add seven community AppSource knowledge articles with paired good and bad AL samples
- cover company-open subscriptions, profile objects, invariant date literals, page action placement, web-service UI calls, searchable entry points, and time-zone offsets
- update the AppSource review skill to route concrete AL signals to the new guidance
- keep the rules focused on AL source and avoid duplicating privacy, style, security, and upgrade domain ownership

## Validation

- `pwsh ./.github/scripts/Test-KnowledgeIndex.ps1 -Root .` (279 articles)
- `pwsh ./tools/Test-ReviewFixtures.ps1 -Root .` (32 cases across 16 leaf domains)
- `git diff --check`
- editor diagnostics


NKarolak (changes requested):
Please mind the rebranding half a year (or longer) ago: "AppSource extension" --> "Marketplace extension".

demiliani:
> Please mind the rebranding half a year (or longer) ago: "AppSource extension" --> "Marketplace extension".

I know (thinked on that yesterday), but then the name should be changed on all the Microsoft's folder rules too. However I don't think this is a need for AI ;)

JesperSchulz (changes requested):
Most of these rules map cleanly to the Marketplace technical-validation checklist, but three correctness issues remain.

1. **`Message` is described as a callback failure, but it has different runtime behavior.** `keep-web-service-paths-free-of-ui-calls.md` says calls “including `Message`” can terminate the service request. Microsoft’s web-service/background-session behavior documents `Message` as suppressed and logged; `Confirm`, `StrMenu`, modal pages, uploads/downloads, etc. are the callback-failure cases. Marketplace validation can still reject message boxes, and a message is ineffective as a service response, but the article and reviewer cue should give that reason rather than teach agents that `Message` throws `NavNCLCallbackNotAllowedException`.

2. **The company-open bad fixture does not model the forbidden subscription correctly.** `do-not-subscribe-to-company-open-events.bad.al` targets `Codeunit::"System Initialization"`, but that codeunit publishes `OnAfterLogin`, not `OnAfterCompanyOpen`. Use a valid `OnBeforeCompanyOpen`/`OnAfterCompanyOpen` subscription shape for the BC versions this rule covers (historically the company-open/ApplicationManagement publisher), so the anti-pattern is compilable and AS0061-relevant rather than failing because the event does not exist on the named object.

3. **The web-service bad fixture contains two unrelated localization defects.** Its hard-coded `Confirm` and `Message` format strings trigger the style rule for Label-backed user text, as the specialized review confirmed. Use `Qst`/`Msg` Labels so the sample isolates only the web-service UI defect and does not teach a second bad pattern.

Please also adopt the reviewer’s terminology feedback in the seven new articles: say **Marketplace extension/app** in prose. Keep `AppSourceCop`, the `appsource` domain/path, and skill identifiers unchanged because those are product/tool and repository contract names; this does not require a repo-wide rename.

The profile-object, invariant-date, company-open prohibition, action-placement, UsageCategory intent guard, and time-zone guidance otherwise look sound. All repository validators pass.

community/knowledge/appsource/use-invariant-date-literals.good.al
```diff
@@ -0,0 +1,7 @@
+codeunit 50100 "Rental Period Defaults"
+{
+    procedure GetPolicyStartDate(): Date
+    begin
+        exit(20250131D);
```
JesperSchulz, line 5: The paired fixtures do not preserve behavior: the bad sample parses `01/02/2025`, while this good sample returns January 31, 2025. Because the review skill may adapt a `.good.al` companion into suggested code, an agent could “fix” the format defect while silently changing the business date. Please make both samples represent the same date—for example, `'01/31/2025'` here’s locale-dependent counterpart versus `20250131D` in this fixture.

community/knowledge/appsource/do-not-subscribe-to-company-open-events.bad.al
```diff
@@ -0,0 +1,13 @@
+codeunit 50100 "Rental Company Open"
+{
+    [EventSubscriber(ObjectType::Codeunit, Codeunit::"Application Management", OnAfterCompanyOpen, '', false, false)]
```
JesperSchulz: This still is not valid for the article's declared `[all]` scope: `Application Management` was removed in BC v14, so the fixture cannot compile against modern symbols. The exact forbidden subscription is also already reported deterministically by AppSourceCop AS0061, while current `main` excludes mechanical compiler/analyzer duplicates from BCQuality knowledge. Please resolve this during the rebase by removing this article, companions, and routing, or demonstrate genuinely additional contextual guidance with accurate version metadata and a publisher valid for that narrowed range.

JesperSchulz (changes requested):
The earlier terminology, web-service behavior, localization, and event-publisher concerns were addressed, and all checks pass at this head. Two correctness issues remain inline.

The PR is also currently conflicting with `main` in `al-appsource-review.md`. Please rebase and preserve the newer standalone `folder-path` flow and the current compiler/analyzer boundary when resolving it; the green checks do not validate the eventual merged tree.

The rest of the guidance remains useful, but the fixtures must preserve behavior and the company-open rule must align with the repository's current analyzer-duplication policy before these articles become authoritative agent guidance.

JesperSchulz (changes requested):
The previous date-fixture, company-open/AppSourceCop duplication, terminology, web-service, localization, publisher, and merge-conflict concerns are resolved. The merged tree matches this head, and the practical local validators pass.

One small but consequential routing contradiction remains. The expanded Worklist now intentionally applies to source-only AL changes such as formatted-string `Evaluate` calls, service-path UI, action placement, `UsageCategory`, profile provisioning, and fixed time-zone offsets. However, the outcome definition still says `not-applicable` when “the diff touches no AppSource permission or app-metadata surface.” A source-only diff can therefore enter the Worklist under the new rules and simultaneously qualify as `not-applicable` at line 84.

Please align that outcome with the expanded applicability rule—for example, define `not-applicable` as no Marketplace-related source, permission, or metadata surface. This affects whether agents run any of the new source checks, so I do not think we should approve until the contract is internally consistent.

JesperSchulz (approved):
Re-reviewed `2ba67ce`. The source-only applicability outcome contradiction is resolved, all earlier correctness issues remain resolved, the exact merged tree matches the reviewed head, and all current checks and practical validators pass. Approving.

PR #145, opened 2026-08-28
Process Context via manual event subscriber pattern

pri-kise:
Adds one community knowledge article with `.good.al` / `.bad.al` samples: how an app driving a process over shared code (a base application report, a posting routine, any object other extensions also hook into) lets those extensions know its process is running.

The reflex answer is a `SingleInstance` boolean, but that state is outside the database transaction — a failed run rolls back the writes and leaves the flag `true` until the company is closed. A manual binding is the safe carrier: its lifetime is tied to a variable's scope, so an error unwinds the stack and takes the binding with it. The documented shape is a public query procedure backed by a `local [InternalEvent]`, answered by an `Access = Internal` / `EventSubscriberInstance = Manual` flag codeunit bound for exactly the run. The public query is the point — a private binding leaves every other extension guessing.

Anti Pattern covers both wrong shapes plus a reviewer note: `BindSubscription` without `UnbindSubscription` is the mechanism here, not a leak. Knowledge-index check passes; 40 lines, no fenced blocks, both samples referenced.

This knowledge is inspired by the current existing logics:
see:
- ZUGFeRD report integratin logic https://github.com/microsoft/ALAppExtensions/pull/29255
- PEPPOL DE Context: https://github.com/microsoft/BCApps/blob/main/src/Apps/DE/PEPPOLDE/app/src/PEPPOL30DEContext.Codeunit.al

### DIFF
--- a/community/knowledge/events/expose-process-context-via-manually-bound-flag.bad.al
+++ b/community/knowledge/events/expose-process-context-via-manually-bound-flag.bad.al
@@ -27,7 +27,6 @@ codeunit 50546 "Process Driver Bad Sample"
     begin
         ProcessState.SetProcessRunning(true);
         RunSharedCode(DocumentNo);
-        FinishProcess(DocumentNo);
         // An error above never reaches this line. The database writes roll
         // back, the single-instance variable does not: ProcessRunning stays
         // true until the company is closed, so every later run in this session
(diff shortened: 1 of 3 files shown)


JesperSchulz (approved):
This is a well-scoped, documented use of manual subscriber lifetime. The public query, module-private InternalEvent, internal manual subscriber, local binding lifetime, nesting behavior, and cross-session limitation are all represented accurately.

PR #147, opened 2026-08-28
ShowMandatory + OnQueryClosePage Check

pri-kise:
## Summary

Two community UI articles with paired AL samples, both encoding a defect reviewers keep finding in real PRs that an LLM does not flag on its own.

- `showmandatory-on-code-required-page-fields.md` — `ShowMandatory` draws the asterisk and enforces nothing, `TestField`/`NotBlank` enforce and render nothing, so a required field ships looking optional. Worst on a `DelayedInsert = true` list page, where the error only arrives once the user leaves the row. Generalizes the review finding on `Job Assigned Resources` in microsoft/BCApps#9315.
- `validate-request-page-input-in-onqueryclosepage.md` — `OnPreReport` runs after the request page has closed, so an error there costs the user everything they typed. Validate in `OnQueryClosePage` too, guarded on `CloseAction`, and keep the check in `OnPreReport` for runs with no request page. The `bad.al` also shows the mirror defect: an unguarded trigger that errors on Cancel and traps the user.


JesperSchulz (changes requested):
The request-page article is sound, but the ShowMandatory article currently contradicts its own sample and can create false positives.

`NotBlank = true` on a primary-key field can already supply the red asterisk. The good sample relies on exactly that for `Code`, yet the Anti Pattern’s reviewer signal says any `NotBlank = true` table field without `ShowMandatory` should be flagged. Please restrict that signal to non-primary-key fields and avoid the earlier blanket statement that `NotBlank` changes nothing about rendering.

Also require evidence that the corresponding page field is visible/editable user input and that the commit path actually requires the user to supply it. A `TestField`/`Error` in any `OnValidate` or `OnModify` does not prove that: the value may be populated by code, non-editable, or conditionally required for another action. The conditional requirement must be mirrored by `ShowMandatory`, as the Best Practice already says.

pri-kise:
How do you come to the conlcusion "NotBlank = true on a primary-key field can already supply the red asterisk."?

Claude said this, but I couldn't verify this behaviour in the client.
see. 
e.g. Default Dimensions
image
```al
        field(3; "Dimension Code"; Code[20])
        {
            Caption = 'Dimension Code';
            NotBlank = true;
            TableRelation = Dimension;

            trigger OnValidate()
            begin
                CheckDimension("Dimension Code");
                UpdateDimensionId();
                if "Dimension Code" <> xRec."Dimension Code" then
                    Validate("Dimension Value Code", '');
            end;
        }
```

there is no asterisk.

Or the salesperson of a team?
image
https://github.com/microsoft/BCApps/blob/main/src/Layers/W1/BaseApp/CRM/Team/SalespersonTeams.Page.al

I've found this part in the docs:
https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/properties/devenv-showmandatory-property
> The NotBlank property can be used to mark the UI on primary key fields with a red asterisk

but this is currently wrong.
I don't know if this is platform bug or intentional.

JesperSchulz:
You're right to challenge that conclusion. I derived it from the Microsoft Learn wording that `NotBlank` “can be used to mark the UI on primary key fields with a red asterisk,” but your current-client examples show that this is not reliable behavior—both fields are members of composite primary keys and still render without an asterisk. I should not have stated that it automatically supplies the marker.

That changes the requested fix as follows:

- Do **not** restrict the detection signal to non-primary-key fields. The article's stronger point—that `ShowMandatory` is the only marker an author should deliberately rely on—is justified by the observed behavior.
- Instead, add `ShowMandatory = true` to the `Code` field in the **good** sample too. It is `NotBlank`, is required before insert, and currently lacks the property that the article says every such page field should carry.
- Please soften “`NotBlank` ... changes nothing about how the field renders” to acknowledge the documented primary-key marking possibility while stating it is not reliable enough to substitute for explicit `ShowMandatory`.

The other false-positive boundary from my review still stands: require evidence that the page field is visible/editable user input and that the relevant commit/action path actually requires the user to supply it. A `TestField`/`Error` in any `OnValidate` or `OnModify` is not sufficient by itself.

Thanks for bringing concrete client evidence; this corrects my first conclusion while preserving the actual issue in the sample and detection rule.

JesperSchulz (approved):
The clarified guidance now matches both Microsoft’s documented `NotBlank` possibility and the observed client behavior: explicit `ShowMandatory` is used in the good sample, and findings require a visible, editable control on a path where the user must supply the value. The request-page guidance remains sound.

PR #150, opened 2026-09-02
Simplify standalone AL code review skill

JesperSchulz:
## Summary

- Rename the standalone plugin skill from `bcquality-al-review` to `al-code-review`.
- Reduce the host-facing `SKILL.md` to a thin adapter that preserves the caller's goal and delegates index preparation, routing, execution, and output semantics to Entry and the dispatched action skills.
- Point plugin manifests at the canonical `skills/` directory and bump the plugin version to `0.2.0`.
- Document the distinction between the host adapter, Entry, the internal `microsoft/skills/review/al-code-review.md` coordinator, and its domain leaves.
- Extend the existing validator to enforce host-skill naming and required frontmatter.

## Rationale

Standalone installation still requires a host-native `SKILL.md`, but it should not become a second review implementation. The adapter now owns only translation into Entry's task context. The remaining reference chain is intentional progressive disclosure rather than duplicated policy.

`al-code-review` also avoids colliding with BC-ALAgents' separately installed `al-review` skill while using the same semantic name as BCQuality's internal coordinator. Their paths and formats distinguish their roles.

## Migration

Explicit invocations and allowlists using `bcquality-al-review` must use `al-code-review` after updating to plugin version `0.2.0`.

## Validation

- `python .github/scripts/validate_frontmatter.py --root .`
- `pwsh -NoProfile -File tools/Test-ReviewFixtures.ps1`
- Live Copilot CLI plugin load with both BCQuality `al-code-review` and BC-ALAgents `al-review` present


gggdttt (approved):

JesperSchulz:
Pushed 87da853 to this branch covering two gaps in the delegation. Both come from the same root cause: `entry.md` is clone-model prose, and the thinned adapter delegates to it wholesale without naming where the plugin model diverges.

**1. Index preparation never ran under a plugin install.** `knowledge-index.json` is gitignored, so a fresh install has none and Preparation must build it. Preparation says to run `pwsh ./tools/Build-KnowledgeIndex.ps1` *from the checkout root*, but a plugin host's working directory is the user's own project. Verified: the relative path does not resolve from a foreign cwd, so nothing is built and READ silently degrades to path-based discovery — the one accelerator the plugin path most needs. The previous adapter avoided this with an explicit `PLUGIN_ROOT`-anchored invocation, which the rewrite dropped.

The fix keeps the adapter thin, since path translation is precisely an adapter's job: step 3 now resolves Entry's repo-relative paths against `PLUGIN_ROOT` and names the absolute index build. The generator defaults its root to `\/..`, so an absolute invocation indexes and writes the correct tree from any directory — verified from `%TEMP%`. The best-effort fallback is restated so a missing `pwsh` does not fail the review.

**2. The layer-pruning caveat was deleted rather than relocated.** The old adapter warned that a plugin ships the whole tree, so `enabled-layers` narrows discovery only — "a selection filter, not a hard security boundary". That text is gone from every file, while Entry's Preparation asserts the opposite as a safety property: "Because a consumer prunes its clone to policy *before* the agent runs [...] never an article the consumer denied". That guarantee does not hold for an install that ships everything.

Entry now scopes the guarantee to consumers that actually prune and states plainly that an unpruned installation gets no deny guarantee from the step. The full caveat is restored in the adapter and summarized in the README plugin section.

**Heads-up on approvals:** this touches `skills/entry.md`, and `skills/README.md` says the contracts "are stable. Changes require a PR approved by both maintainers." The Entry edit is deliberately narrow — it scopes an over-broad claim and adds a path-resolution note, changing no routing, dispatch, or output semantics — but it does move this PR into that category.

The rest of the PR stands: thinning the adapter is the right call, `disabled-skills` and `enabled-layers` are both genuinely present in `entry.md`, the validator addition classifies `skills/*/SKILL.md` correctly and passes clean, and the only remaining `bcquality-al-review` reference is the intentional migration note.

Note this branch is still based on 35a7e72 and is behind `main` (#144 and #149 have since merged); worth a rebase before merging.

PR #156, opened 2026-09-04
18 AL/BC patterns: style, data-modeling, web-services, appsource, breaking-changes, performance, testing

MichaelDieringer:
## Summary

18 knowledge articles generalized from patterns observed across real AppSource/PTE Business Central development, contributed by CURABIS ApS. Each is a self-contained article following the knowledge file format (6-key frontmatter, Description/Best Practice/Anti Pattern, sibling `.good.al`/`.bad.al` samples, no fenced code in the .md).

Targeted at existing Microsoft-owned domains per the README's contribution guidance ("use `/microsoft/knowledge/` for Microsoft-owned domains") rather than `/community/`, since these all fall under domains BCQuality already owns action skills for.

- **style** (7): AL identifier/casing/typing conventions — Boolean vs Option for binary choices, Enum vs Integer for fixed choice sets, modern casing for intrinsic functions, addressable var parameters, namespace verification, English-only identifiers, keeping build output out of the project root.
- **data-modeling** (4): DimensionManagement wiring for new master/document tables, the Check Line / Post Line / Post Batch posting-routine split, primary key immutability post-publish, TestField-before-use for required setup fields.
- **web-services** (3): CalcFields for FlowFields on API pages, keeping consumer-provided key fields editable on insert, recalculating stored/derived fields before exposing them via API.
- **appsource** (1): File data type direct I/O is OnPrem-only.
- **breaking-changes** (1): Prefer the Email module over the obsolete Codeunit 397 "Mail".
- **performance** (1): Document reports should default to Word layout.
- **testing** (1): Randomized, complete test data via library helpers.

Checked against the current `microsoft/knowledge/` corpus before opening this PR (post the recent community→microsoft consolidation) — two originally-drafted articles were dropped as duplicates of existing files (`xrec-is-a-before-image-only-in-some-triggers.md`, `use-library-codeunits-for-test-fixtures.md` + `asserterror-needs-expectederror-and-code.md`).

## Test plan


microsoft/knowledge/data-modeling/do-not-change-primary-key.md
```diff
@@ -0,0 +1,28 @@
+---
+bc-version: [all]
+domain: data-modeling
+keywords: [primary-key, clustered-key, table-design, appsource, breaking-change, schema-upgrade]
+technologies: [al]
+countries: [w1]
+application-area: [all]
+---
+
+# Never Change a Published Table's Primary or Clustered Key Field List
```
JesperSchulz, line 10: This article's companion files are named `do-not-change-primary-key.good.al` / `.bad.al`, so the stem does not match this Markdown file. The frontmatter validator reports both as R14 orphan samples. The same mismatch exists for `testfield-required-setup-field-before-use` versus `testfield-required-setup-field.*`, and `al-identifiers-must-be-english` versus `al-identifiers-english.*`. Please rename either each article or its samples so all three pairs validate.

microsoft/knowledge/style/al-build-output-must-not-pollute-project-root.md
```diff
@@ -0,0 +1,24 @@
...
+application-area: [all]
+---
+
+# Keep AL Build Output Out of the Project Root
+
+> Contributions welcome — open a PR to refine or extend this article.
+
+## Description
+
+When an AL project is built, the compiled `.app` file is placed in the project root by default. Over successive builds, multiple `.app` files accumulate there (e.g. one per version). The AL language server, both in the editor and in build tooling, scans the project folder for symbol packages and can load these compiled artefacts alongside the live source files, which produces `AL0197` duplicate-object errors for every object in the project — with messages that point at source lines rather than at the packaged artefact that is the actual duplicate. The errors are not real; they disappear as soon as the stale `.app` files are removed from the root.
```
JesperSchulz: This contradicts the documented default build behavior without establishing the alleged scanner behavior. Microsoft documents the project folder as the default `.app` output folder; normal root output must therefore be supported. AL0197 means a duplicate object in the extension or a declared dependency, not that the language server loads every `.app` sitting beside the source. Please provide an authoritative/reproducible configuration in which this occurs and scope the rule to it, or remove this causal claim.

microsoft/knowledge/style/al-build-output-must-not-pollute-project-root.md
```diff
@@ -0,0 +1,24 @@
...
+
+## Best Practice
+
+Configure the build output path to a dedicated subfolder that is excluded from language server scanning — for example by setting `al.outputPath` to a folder such as `.output` in `.vscode/settings.json`, or by passing an explicit output path to the build tool being used — and add that folder to `.gitignore`. Before treating an `AL0197` "already declared" error as a source code problem, check the project root for stale `.app` files first; adding root `.app` files to `.gitignore` instead of relocating the output path only hides the accumulation rather than fixing it.
```
JesperSchulz: `al.outputPath` is not a documented AL Language setting. Current AL Language configuration exposes the build destination as `al.compilationOptions.outFolder`. As written, agents will generate a setting that the extension does not consume.

microsoft/knowledge/breaking-changes/prefer-email-module.md
```diff
@@ -0,0 +1,28 @@
...
+application-area: [all]
+---
+
+# Send email through the Email module, not Codeunit Mail (397)
+
+> Contributions welcome — open a PR to refine or extend this article.
+
+## Description
+
+Older AL code sends email by calling `Codeunit Mail (397)`. Business Central's current extensibility model is a different, richer object set — `Codeunit Email`, table `Email Message`, `enum "Email Scenario"`, and the `Email Account`/`Email Connector` interface (Microsoft 365, Current User, SMTP, or a custom connector). New code built on `Codeunit Mail` inherits its SMTP-era, single-connector assumptions and leaves no Sent/Outbox trail behind.
```
JesperSchulz: `Email Message` is codeunit 8904 in `System.Email`, not a table. The good fixture declares it correctly as `Codeunit "Email Message"`, so both occurrences in the article should be corrected. Please also distinguish that codeunit from the underlying outbox/sent-email storage rather than promising a table record from this variable.

microsoft/knowledge/appsource/file-datatype-saas.md
```diff
@@ -0,0 +1,28 @@
...
+application-area: [all]
+---
+
+# The File data type's direct I/O methods are OnPrem-only
+
+> Contributions welcome — open a PR to refine or extend this article.
+
+## Description
+
+The classic `File` variable type — `Open`/`Create`/`Read`/`Write`/`Close` against a path on the local or server filesystem — only works on-premises, because there is no accessible filesystem in the SaaS/cloud sandbox. Any extension meant to run in Business Central Online must not rely on `File.Open`, `File.Create`, `File.Read`, or `File.Write` for its core functionality: code built this way compiles but fails, or is silently skipped, in the cloud.
```
JesperSchulz: The failure mode is wrong. The filesystem methods are OnPrem-scoped, so a Cloud-targeted extension that calls `File.Open`/`Create`/`Read`/`Write` fails compilation; it does not successfully compile and then fail or become "silently skipped" in SaaS. Please state the compile-time restriction, while separately explaining that server filesystem paths are unavailable online.

microsoft/knowledge/data-modeling/dimension-management-wiring.good.al
```diff
@@ -0,0 +1,28 @@
...
+        {
+            CaptionClass = '1,1,1';
+            TableRelation = "Dimension Value".Code where(
+                "Global Dimension No." = const(1), Blocked = const(false));
+
+            trigger OnValidate()
+            var
+                DimMgt: Codeunit DimensionManagement;
+            begin
+                DimMgt.ValidateShortcutDimCode(1, "Global Dimension 1 Code");
```
JesperSchulz: This calls an API that does not exist on current codeunit 408 `DimensionManagement`. Standard master tables define their own `ValidateShortcutDimCode` wrapper and call `DimMgt.ValidateDimValueCode(FieldNumber, ShortcutDimCode)` followed by `SaveDefaultDim`; document/journal tables call `ValidateShortcutDimValues(..., "Dimension Set ID")`. Please make the article distinguish those two models and make the good fixture use a real current signature.

microsoft/knowledge/data-modeling/do-not-change-primary-key.good.al
```diff
@@ -0,0 +1,24 @@
...
+table 50101 "Period Stats By Flow"
+{
+    fields
+    {
+        field(1; "Period Start"; Date) { }
+        field(2; Flow; Enum "Some Flow") { }
+    }
+    keys
+    {
+        key(PK; "Period Start") { Clustered = true; }
```
JesperSchulz: The replacement table does not actually model the new discriminating dimension: because `Flow` is absent from its primary key, it can store only one row per period and cannot store one row per `(Flow, Period Start)`. The new table's own key should include `Flow`; only the already-published table's key must remain frozen.

microsoft/knowledge/style/namespace-must-be-verified-from-source.md
```diff
@@ -0,0 +1,28 @@
...
+application-area: [all]
+---
+
+# Verify a namespace from the object's own source file, never by inference
+
+> Contributions welcome — open a PR to refine or extend this article.
+
+## Description
+
+Since Business Central 2024 release wave 1, Microsoft's own objects are organized under a deep `Microsoft.*` namespace tree that has been renamed and restructured repeatedly. Guessing a namespace from an object's name, from an older codebase, or from general familiarity produces a `using` statement that can look plausible, compile in isolation, and still resolve to the wrong object or fail in the AL Language Server that VS Code actually uses to report errors. The only reliable source for an object's namespace is line one of that object's own source file.
```
JesperSchulz: Source-file line 1 is not the only reliable namespace source. Dependency symbol packages and AL symbol tooling are authoritative and are often the only available source for closed-source dependencies; a source file may also have comments/directives before its `namespace` declaration. The sound rule is "verify from source or symbols; do not guess," not this line-one-only contract.

microsoft/knowledge/testing/test-data-must-be-random-and-complete.md
```diff
@@ -0,0 +1,28 @@
...
+
+> Contributions welcome — open a PR to refine or extend this article.
+
+## Description
+
+An AL test suite should assume an empty database. Test data must be created programmatically inside the test rather than assuming a specific code, number, or name already exists in the environment — a hardcoded lookup against an assumed-existing record makes the test fail for reasons unrelated to the code under test. Every mandatory field on a created record also needs a value that respects its declared length; a partial setup that merely passes validation is not sufficient.
+
+## Best Practice
+
+Use the standard library codeunits (`Library - ERM`, `Library - Inventory`, `Library - Sales`, `Library - Utility`) to create records with collision-free random values, and fill every mandatory field with randomized, correctly-sized data. Reserve hardcoded values for tests that validate an external contract itself — a fixed JSON schema, an EDIFACT message, a counterparty code — where the hardcoded value documents the specification rather than arbitrary test logic.
```
JesperSchulz: These absolutes do not describe BC test environments. Tests commonly run in an initialized company containing system/setup data; they should create the business records they own and avoid arbitrary assumed records, not assume an empty database. Likewise the standard libraries generate valid/unique-enough fixture data through number series and random helpers, but do not promise mathematically collision-free random values. Please reframe both requirements around isolation, ownership, and avoiding hardcoded environmental dependencies.

microsoft/knowledge/style/binary-choice-must-be-boolean.md
```diff
@@ -0,0 +1,28 @@
...
+application-area: [all]
+---
+
+# Binary yes/no choices must be typed as Boolean, not Option or Integer
+
+> Contributions welcome — open a PR to refine or extend this article.
+
+## Description
+
+When a field or variable represents exactly two states — yes/no, on/off, active/inactive, blocked/not blocked — it should be typed `Boolean`. Modeling that same two-state choice as an `Option`/`Enum` with two members, or as an `Integer` with two magic-number values, adds a layer of indirection a reader has to resolve before understanding the code, and it invites a multi-branch check where a simple `if X then` would do. This is distinct from a genuine multi-value choice with more than two named states, which legitimately calls for `Enum` — the line is the state count.
```
JesperSchulz: State count alone is not a safe type rule. A two-member enum can deliberately encode named domain semantics, implement an interface, preserve an existing contract, or allow future enum extension; `Boolean` is appropriate only when the domain is genuinely a stable predicate/flag. Please scope this to true yes/no flags and stop classifying every two-state enum as an anti-pattern.

microsoft/knowledge/performance/document-report-word-layout.md
```diff
@@ -0,0 +1,28 @@
...
+application-area: [all]
+---
+
+# Document reports should default to a Word layout, not RDLC
+
+> Contributions welcome — open a PR to refine or extend this article.
+
+## Description
+
+A document report — an invoice, statement, order confirmation, or any report meant to be printed, emailed, or exported as a single-record document — should default to a `Word` rendering layout rather than `RDLC`. RDLC layouts run in a sandboxed app domain that only lives for the current report invocation, which is slower for UI-related actions such as emailing the resulting document, than a Word layout, which is not subject to that sandbox constraint. This does not apply to every report: tabular/list reports with heavy aggregation or calculated columns are still often a better fit for RDLC or Excel.
```
JesperSchulz: Please substantiate and version-scope this performance guarantee. Current Microsoft guidance describes Word as convenient for document layouts and RDLC/Excel as suitable for other layout needs, but does not establish that every document report should default to Word because RDLC alone incurs a per-invocation sandbox AppDomain cost during email actions. Without an authoritative benchmark/source, this turns a layout-design choice into a universal performance rule agents will enforce.

JesperSchulz (changes requested):
There are useful topics here, but this is not safe to merge as agent guidance yet.

The repository validator reports six R14 errors because three article stems do not match their companion sample stems (`do-not-change-primary-key-field-list`, `testfield-required-setup-field-before-use`, and `al-identifiers-must-be-english`). Beyond that structural failure, several normative claims and fixtures conflict with the current platform/API: `Email Message` is a codeunit, the dimension sample calls a nonexistent `DimensionManagement.ValidateShortcutDimCode` API, cloud-scoped `File` calls fail compilation rather than compiling and being silently skipped, and the proposed AL build setting is not an AL Language setting.

The selection contract also needs updating. None of the leaf review skills changes in this PR, and several new articles cannot enter their worklists from the currently enumerated tokens/cues (for example `file-datatype-saas`, `prefer-email-module`, `namespace-must-be-verified-from-source`, and `var-parameters-require-an-addressable-variable`). The skills currently claim their targeted checks cover every current article, so please add deterministic cues/tokens for the new rules rather than relying on fuzzy topic matching.

I have left inline details on the confirmed factual and fixture issues. The knowledge-index and neutral-fixture validators pass, but frontmatter/sample validation fails. The CLA check is also still pending and must be resolved separately before merge.

microsoft/knowledge/performance/document-report-word-layout.md
MichaelDieringer: Thanks Jesper. I agree that this should not be interpreted as a universal guarantee that Word is faster than RDLC for every workload.

However, the recommendation itself is based directly on current Microsoft documentation rather than on an inferred benchmark. 

Microsoft's current *Report Design Overview* states:

> “RDL layouts can result in slower performance with document reports, regarding actions that are related to the user interface (for example, like sending emails) compared to Word layouts.”

and explicitly continues:

> “When developing layouts for document reports, we recommend that you design Word layouts instead of RDL.”

The documentation also explains the sandbox app-domain behavior as the reason for that recommendation.

The current report troubleshooting guidance repeats that RDL layouts can result in slower performance than Word layouts for document reports, particularly for UI-related actions.

So I agree that “Word is always faster” might be too strong, and I will scope the article accordingly. But I believe retaining “prefer Word for document reports” is consistent with Microsoft's published guidance.

I suggest changing the rule from an unconditional default to:

“For document reports, prefer Word layouts over RDL unless specific layout requirements favor RDL. Microsoft recommends Word layouts for document reports and notes that RDL layouts can result in slower performance for UI-related actions such as emailing.”

Sources:

* Microsoft Learn: Report Design Overview (https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-report-design-overview)
* Microsoft Learn: Creating an RDL layout report (https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-report-performance)
* Microsoft Learn: Troubleshooting reports / Report performance (https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-reports-troubleshooting)

microsoft/knowledge/style/binary-choice-must-be-boolean.md
MichaelDieringer: Agree, good point.

Reading the rule now, the rule as written is way too absolute. A two-value enum is not inherently poor modeling. 

If the domain concept is genuinely a true/false state — enabled/disabled, present/absent, yes/no, then a Boolean is usually the clearer representation.

But there are also valid domain concepts that currently have exactly two named alternatives, for example Inbound/Outbound, Debit/Credit, or Buy/Sell. In those cases, an enum can express the business semantics more clearly than forcing one value to become the implicit negation of the other.

So I think the rule should be narrowed to target Boolean semantics more than simply counting enum values.

I’ll revise it accordingly, for example:

“Use a Boolean for concepts that are inherently true/false states. Do not replace a meaningful two-option domain model with a Boolean solely because the enum currently has two values.”

This makes the rule more precise and avoids agents mechanically converting valid domain enums into Booleans.

microsoft/knowledge/testing/test-data-must-be-random-and-complete.md
MichaelDieringer: I don't think I agree with this one.

The intent of the rule is not that tests should use uncontrolled randomness or produce different, irreproducible scenarios on every run. The point is that tests should create generalized test data instead of depending on hardcoded business records or pre-existing company data.

This is also consistent with Microsoft's own testing guidance. Microsoft recommends using hardcoded values when a specific value is required by the test, and otherwise using generated/random test data.

At the same time, the test framework supports reproducibility through seeded pseudo-random data. Library - Random, SetSeed, and the newer Any library exist precisely so tests can generate generalized data while still reproducing failures consistently.

I think the important distinction is between incidental test data and values that are part of the behavior being tested:

- Tests should be self-contained and create the records and setup they need.
- Incidental values such as record identifiers, names, descriptions, and similar fixture data should generally be generated rather than tied to specific existing data.
- Generated values should be deterministic/reproducible where necessary.
- Values that materially define the scenario should be explicit and fixed. For example, amounts, quantities, percentages, dates, thresholds, rounding precision, or other values where the expected result depends on the exact input should not be randomized merely for the sake of randomization.

A rounding test, for example, should obviously use deliberately chosen values around the rounding boundary. Likewise, a calculation test should use explicit quantities and amounts when those values are part of what the test is asserting.

This is fundamental to making a test suite repeatable across companies, databases, environments, and repeated runs without relying on specific fixture data, while still keeping the business behavior under test explicit and understandable.

There is also another reason why generated test data is useful beyond avoiding dependencies on predefined records: it can exercise the data model itself.

Generated values can expose issues in TransferFields, field mappings, truncation, field-length assumptions, character handling, or copying between records. A test suite that always uses short constants such as "TEST" may never reveal that a value is incorrectly truncated, that a destination field cannot represent the source data correctly, or that assumptions about field definitions are wrong.

So generated data is not just about making tests generic. Used appropriately, it also increases coverage of the data model and the paths by which data moves through the application.

That still does not mean that semantically significant values should be randomized. Amounts, quantities, rounding boundaries, percentages, and similar inputs should remain explicit whenever the expected behavior depends on their exact value.

I would therefore rather clarify “random” as “generated/reproducible test data where the exact value is not material to the scenario” than remove the pattern.

microsoft/knowledge/style/namespace-must-be-verified-from-source.md
MichaelDieringer: The intent here is much narrower than general namespace governance.

This rule is specifically about namespace resolution when an agent creates AL variables or references existing application objects.

I have repeatedly seen AI-generated AL code invent plausible-looking namespaces that simply do not exist. For example, when creating:

var
    Customer: Record Customer;

the agent may also generate a using directive for a guessed namespace based on the object name or functional area.

That is exactly what this rule is intended to prevent.

If the agent needs the namespace for Record Customer, it should inspect the actual Customer table definition available in the referenced source/symbols and use the namespace declared there.

In other words:

Don't guess the namespace. Resolve it from the object you are actually referencing. The same applies to codeunits, enums, interfaces, pages, and other AL objects.
So I would keep the rule, but clarify the wording to make the scope explicit:

When adding a using directive for an existing AL object, resolve the namespace from the object's actual source or symbol definition. Do not infer or invent namespaces from object names, functional areas, or naming conventions.

The purpose is simply to prevent agents from producing syntactically plausible but non-existent namespace references.

microsoft/knowledge/data-modeling/do-not-change-primary-key.good.al
MichaelDieringer: Agree...

The rule itself is still what I intended: the already-published table must keep its existing primary key unchanged, and the new discriminating dimension should be modeled separately.

Our good.al sample is wrong. The new table must of course include Flow in its own primary key; otherwise it can only store one row per Period Start and does not actually model the additional dimension.

I’ll change it to something like:

keys
{
    key(PK; Flow, "Period Start")
    {
        Clustered = true;
    }
}

Thanks for catching that.

microsoft/knowledge/data-modeling/dimension-management-wiring.good.al
MichaelDieringer: Ohh yes - nice catch :-)

The intent of the pattern is still valid, but the good fixture uses the wrong API and the article currently conflates two different dimension models.

For master-data tables, the standard pattern is to expose a local ValidateShortcutDimCode wrapper and have that call the current DimensionManagement APIs for validating the dimension value and persisting the corresponding Default Dimension.

For document and journal tables, the model is different because dimensions are represented through "Dimension Set ID", and validation goes through ValidateShortcutDimValues(...).

So the article should distinguish explicitly between:

Master data / entities: Default Dimension records tied to the entity.
Documents / journals / transactional records: Dimension Set ID and shortcut-dimension validation against that dimension set.

The good example should also use the actual current signatures from codeunit 408 rather than a simplified API that does not exist.

I’ll correct both the fixture and the wording accordingly.

Many thanks, this is an important distinction, especially since the purpose of these patterns is to prevent agents from inventing plausible-looking BC APIs.

microsoft/knowledge/appsource/file-datatype-saas.md
MichaelDieringer: Correct, he failure mode is described incorrectly.

The direct filesystem APIs are OnPrem scoped, so code targeting Business Central Online cannot successfully compile against File.Open, File.Create, File.Read, File.Write, etc. This should be described as a compile-time restriction, separately from the architectural fact that a server/local filesystem path is not available to an extension running in Business Central Online.

I’ll update the wording accordingly.

The underlying guidance remains the same: cloud-compatible code should use stream-based APIs such as UploadIntoStream / DownloadFromStream (or other appropriate cloud-accessible storage APIs) rather than direct filesystem access.

microsoft/knowledge/breaking-changes/prefer-email-module.md
MichaelDieringer: "Email Message" is codeunit 8904 in System.Email, not a table. The good fixture already declares it correctly as:

EmailMessage: Codeunit "Email Message";

I’ll correct both occurrences in the markdown.

You are also right that the article should distinguish the "Email Message" codeunit from the underlying Sent/Outbox/Draft storage rather than implying that the variable itself represents a persisted table record.

Nice done, catching this one. Many thanks.

microsoft/knowledge/testing/test-data-must-be-random-and-complete.md
JesperSchulz: I think we agree on the intended pattern, and I am not asking to remove generated test data. Your proposed distinction—self-contained fixtures, generated/reproducible incidental values, and explicit values when they define the scenario—would resolve the substantive concern.

The two phrases that still need to be removed or rewritten are specifically **“assume an empty database”** and **“collision-free random values.”** A BC test company normally contains initialized system/setup data, so the contract should be independence from unrelated business records rather than emptiness. Likewise, the libraries provide generated/valid values (often made unique through number series or controlled randomness), but not a general collision-free guarantee. With those two absolutes corrected, I’m aligned with the revised direction.

microsoft/knowledge/style/al-build-output-must-not-pollute-project-root.md
MichaelDieringer: I think this one needs to be scoped more carefully.

The intent is not to claim that generating an .app somewhere under the AL project is inherently invalid. The current AL tooling explicitly supports choosing where compilation artifacts are written. For example, ALTool exposes:

--outfolder <path> -> “Output folder for compilation artifacts.”

Microsoft documents this as part of the current ALTool tooling for automated and agent-driven development:
https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-al-tool

Likewise, the current al_build agent tool supports an explicit outputPath for the generated .app package.

The issue I want to prevent is therefore not a particular directory being technically invalid. It is agents leaving build artifacts in arbitrary source locations, or treating generated .app files as part of the source tree.

So I think the rule should be reframed around deliberate artifact handling rather than saying that the project root must never contain build output.

Something along the lines of:

Build artifacts should be written to an intentional output location and must not be treated as source files or committed unintentionally. For automated or agent-driven builds, prefer a dedicated output folder.

microsoft/knowledge/style/al-build-output-must-not-pollute-project-root.md
MichaelDieringer: True. The causal claim around AL0197 is not sufficiently substantiated and must not be in the rule.

The intent was to prevent agents from scattering generated build artifacts through the source tree, but I incorrectly connected that to the compiler resolving arbitrary .app files from the project root.

As you point out, the normal build behavior itself permits output in the project folder, so root-level .app output cannot reasonably be treated as invalid by itself. AL0197 should also not be attributed to an .app merely being present beside the source without a reproducible configuration showing that behavior.

I’ll remove that causal claim and reframe the pattern as repository/build-artifact hygiene rather than compiler correctness:

Build artifacts should be written to an intentional output location and should not be treated as source files or committed unintentionally. For automated or agent-driven builds, prefer a dedicated output folder.

That should preserve the actual intent without claiming behavior that the AL tooling does not document.

microsoft/knowledge/data-modeling/do-not-change-primary-key.md
MichaelDieringer: agree - I’ll rename them so the stems match and the R14 validator can associate the samples correctly.

While working, I’ll fix the same mismatch for the testfield-required-setup-field-before-use and al-identifiers-must-be-english patterns as well.

MichaelDieringer:

JesperSchulz (commented):
Follow-up after the latest two commits: many of the original issues are fixed, including fixture discovery, unsupported settings guidance, the Email Message object type, cloud File behavior, the replacement-table key, Boolean semantics, and Word/RDL guidance. The remaining correctness issues are inline below.

One integration issue also remains: please add deterministic candidate-selection cues for the new data-modeling, testing, performance, and web-services rules whose actual anti-patterns are not already discoverable by their owning review skills. Passing the content validators does not by itself make those rules reachable by an agent review.

microsoft/knowledge/data-modeling/dimension-management-wiring.md
```diff
@@ -0,0 +1,28 @@
...
+application-area: [all]
+---
+
+# Wire dimension support through DimensionManagement, not ad hoc fields
+
+> Contributions welcome — open a PR to refine or extend this article.
+
+## Description
+
+Adding dimension support to a custom master or document table is not just a matter of adding a `Code[20]` field. Business Central expects a specific set of hooks into `Codeunit "Dimension Management"` so a dimension value is validated, persisted as a Default Dimension record, and flows through to transactions the same way it does for every standard table. Skipping any one hook produces a field that looks correct in the designer but silently fails to save, validate, or carry through to postings.
```
JesperSchulz: This still conflates two different dimension models. Master data persists **Default Dimension** records, while transactional records carry a **Dimension Set ID** assembled from inherited/default dimensions. A master record does not generally acquire a DimSetID through `SaveDefaultDim`, and a document needs more than adding shortcut fields. Please split these cases and make the good fixture implement every hook the prose requires, including inheritance into the document's dimension set.

microsoft/knowledge/web-services/api-page-flowfields-must-be-calcfields.md
```diff
@@ -0,0 +1,28 @@
...
+application-area: [all]
+---
+
+# Explicitly Calculate FlowFields on API Pages
+
+> Contributions welcome — open a PR to refine or extend this article.
+
+## Description
+
+FlowFields are not stored in the database — Business Central computes them on demand. Regular pages trigger that calculation automatically while rendering, but API pages do not. A FlowField referenced in an API page's layout returns an empty value to external consumers unless it is calculated explicitly, producing a silent data gap in OData responses that is easy to miss in review.
```
JesperSchulz: This premise is not supported by the platform documentation. A FlowField used directly as a page control's source is automatically calculated, and the documentation does not carve out API pages: https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-flowfields. Please provide a current reproducible API-specific case and narrow the rule to it, or remove this article; as written it will make agents add redundant `CalcFields` calls everywhere.

microsoft/knowledge/breaking-changes/prefer-email-module.bad.al
```diff
@@ -0,0 +1,13 @@
+codeunit 50103 "Order Confirmation Notifier"
+{
+    procedure Send(FromName: Text; ToAddress: Text; Subject: Text; Body: Text)
+    var
+        Mail: Codeunit Mail;
+        MailSent: Boolean;
+    begin
+        Mail.CreateMessage(FromName, ToAddress, '', Subject, Body, true);
```
JesperSchulz: This legacy example is not valid against the current `Mail` codeunit: `CreateMessage` has a different signature, and codeunit 397 no longer exposes the `Send` or `GetErrorDesc` calls used below. Because the article is `[all]`, the bad fixture still needs to be valid AL for a clearly identified source version, or the article needs explicit version scoping and a compilable migration example.

microsoft/knowledge/data-modeling/check-post-line-batch-pattern.md
```diff
@@ -0,0 +1,28 @@
...
+application-area: [all]
+---
+
+# Split posting routines into Check Line / Post Line / Post Batch
+
+> Contributions welcome — open a PR to refine or extend this article.
+
+## Description
+
+Every journal-based posting routine in Business Central is split across three companion codeunits with distinct, non-overlapping responsibilities: `Check Line` validates one line, `Post Line` writes exactly one line to the ledger, and `Post Batch` loops both across the journal. A document posting routine (posting one document at a time) calls `Post Line` directly and skips `Post Batch`. A new posting routine that blurs this split either misses functionality other code expects to call directly, or exposes an interaction surface it shouldn't.
```
JesperSchulz: This turns a useful standard architecture into a universal invariant. Not every journal-based posting implementation has exactly these three companions or these non-overlapping responsibilities, and document posting does not universally call the same `Post Line` directly. Please present this as a pattern to evaluate in context, identify the standard implementations it is based on, and replace the sample's nonexistent `"Meter Jnl.-Check Line"` dependency so the good fixture is internally complete.

microsoft/knowledge/testing/test-data-must-be-random-and-complete.good.al
```diff
@@ -0,0 +1,13 @@
+[Test]
+procedure PostsSalesOrderForRandomCustomer()
+var
+    Customer: Record Customer;
+    SalesHeader: Record "Sales Header";
+begin
+    // Random, collision-free customer — no assumption about what exists.
```
JesperSchulz: The prose correctly stopped promising uniqueness, but the good fixture still calls this “collision-free.” `LibrarySales.CreateCustomer` greatly reduces collision risk and handles normal test setup; it does not establish a mathematical collision-free guarantee. Please keep the fixture aligned with the corrected article wording.

microsoft/knowledge/style/fixed-choice-set-must-use-enum-not-integer.md
```diff
@@ -0,0 +1,28 @@
...
+application-area: [all]
+---
+
+# A fixed set of named choices must use Enum, not a raw Integer
+
+> Contributions welcome — open a PR to refine or extend this article.
+
+## Description
+
+When a variable or field can only take on a fixed set of more than two named, mutually exclusive states — a difficulty level, a document type, a processing status — it should be typed as `Enum` (or `Option` when extending an object that still uses the legacy type). Representing that same state as a plain `Integer` and tracking the meaning of each value in a comment or in a developer's head is a magic-number anti-pattern: the compiler cannot catch an out-of-range value, and branches read as opaque numbers instead of names. This is distinct from a true two-state choice, which should be `Boolean` rather than an enumeration — the line is the state count.
```
JesperSchulz: “The line is the state count” reintroduces the same incorrect heuristic that the revised Boolean article removed. Two named states are not automatically Boolean: a two-value business status can still need enum semantics, extensibility, captions, or a meaningful third state later, while Boolean is appropriate for an intrinsically binary predicate. Please make this distinction semantics-based here too.

microsoft/knowledge/style/namespace-must-be-verified-from-source.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Best Practice
+
+When referencing an existing AL object, resolve its namespace from that object's actual source file or symbol definition — never infer or invent one from its name, functional area, or naming convention.
+
+See sample: `namespace-must-be-verified-from-source.good.al`.
+
+## Anti Pattern
+
+Writing a `using` statement from memory, from an incomplete path, or from a plausible-looking guess. It can compile in one build environment while still failing to resolve in VS Code, because the two use different namespace resolution.
```
JesperSchulz: The source/symbol guidance is now sound, but this explanation is not. The compiler and AL language server do not intentionally use different namespace-resolution rules; differences usually come from stale/different symbols, configuration, or build context. Please describe those concrete causes rather than teaching agents that the two resolvers inherently disagree.

microsoft/knowledge/style/intrinsic-al-functions-must-use-modern-casing.md
```diff
@@ -0,0 +1,28 @@
...
+application-area: [all]
+---
+
+# Intrinsic AL function calls use modern casing, not legacy ALL-CAPS
+
+> Contributions welcome — open a PR to refine or extend this article.
+
+## Description
+
+AL is case-insensitive, so `MESSAGE(...)`, `ERROR(...)`, `CONFIRM(...)`, and `STRSUBSTNO(...)` compile and run identically to `Message(...)`, `Error(...)`, `Confirm(...)`, and `StrSubstNo(...)`. Modern AL — the VS Code tooling's default formatter output, Microsoft's own current samples, and current reference codebases — writes intrinsic/built-in function calls in the casing Microsoft assigns to the function's declared name, typically PascalCase. ALL-CAPS calls are a holdover from classic C/AL and signal code that has not been modernized, even though it compiles and runs correctly. Reserved keywords such as `if`, `begin`, and `for` are a separate, already-tooled concern; intrinsic function names are identifiers, not keywords, so that tooling does not catch ALL-CAPS intrinsic function calls.
```
JesperSchulz: Please remove the claim that this is the VS Code formatter's default output unless it can be demonstrated. The formatter preserves or normalizes only particular syntax and should not be relied on to recase every intrinsic identifier. The Microsoft/current-codebase convention is sufficient justification without assigning behavior to the formatter.

microsoft/knowledge/data-modeling/dimension-management-wiring.md
MichaelDieringer: Fixed in 584143b — split the article into the two distinct models: master data (Default Dimension via `ValidateDimValueCode`/`SaveDefaultDim`, no `Dimension Set ID` field) versus transactional/document data (a single `Dimension Set ID` built from `AddDimSource` + `GetDefaultDimID`, inheriting the master's Default Dimension records, with `ValidateShortcutDimValues` updating that same ID rather than writing a separate Default Dimension record). Added a second, document-table example (`Course Registration Header`) alongside the existing master-table one, using the real 3-arg `AddDimSource(var DefaultDimSource, TableId, No)` and 6-arg `GetDefaultDimID(DefaultDimSource, SourceCode, var GlobalDim1Code, var GlobalDim2Code, InheritFromDimSetID, InheritFromTableNo)` signatures — verified against `DimensionManagement.Codeunit.al` and the real call site in `ExchRateAdjmtProcess.Codeunit.al`.

microsoft/knowledge/web-services/api-page-flowfields-must-be-calcfields.md
MichaelDieringer: Agreed — deleted the article and both fixtures in 584143b. Couldn't reproduce an API-page-specific exception to the documented automatic-calculation behavior, so there's no article here.

microsoft/knowledge/breaking-changes/prefer-email-module.bad.al
MichaelDieringer: Fixed in 584143b. `Codeunit Mail` has no `Send`/`GetErrorDesc` members — verified against `Mail.Codeunit.al`. The bad fixture now calls the real 7-argument `CreateMessage(ToAddresses, CcAddresses, BccAddresses, Subject, Body, ShowNewMailDialogOnSend, RunModal)`. Also corrected the prose: `CreateMessage`'s own current implementation only raises `OnBeforeCreateMessage`/`OnCreateMessageOnBeforeClearAttachmentFileNames` — it doesn't send anything by itself — so "it still runs" is replaced with language about depending on whatever compatibility subscriber happens to be wired up.

microsoft/knowledge/data-modeling/check-post-line-batch-pattern.md
MichaelDieringer: Fixed in 584143b — reframed from a universal invariant to "the standard shape to evaluate against," naming the real codeunits it's based on (Gen./Item/Resource/Job/FA/Insurance/CA Jnl.-Check Line/-Post Line/-Post Batch — all verified to exist under those names). Also added the missing `Check Line` companion codeunit to the good fixture so it's internally complete.

microsoft/knowledge/testing/test-data-must-be-random-and-complete.good.al
MichaelDieringer: Fixed in 584143b — removed the leftover "collision-free" wording from the fixture comment; it now reads "freshly created customer, owned by this test," matching the already-corrected article text.

microsoft/knowledge/style/fixed-choice-set-must-use-enum-not-integer.md
MichaelDieringer: Fixed in 584143b — removed "the line is the state count" and replaced it with the same semantics-based distinction now in `binary-choice-must-be-boolean.md`: a two-value Enum with named alternatives stays an Enum, Boolean is for a true/false predicate regardless of member count.

microsoft/knowledge/style/namespace-must-be-verified-from-source.md
MichaelDieringer: Fixed in 584143b — removed the claim that the compiler and AL Language Server apply different namespace-resolution rules; the Anti Pattern section now attributes a wrong resolution to stale/mismatched symbols, a different build configuration, or the object's actual current source instead.

MichaelDieringer:
@JesperSchulz Addressed this round in 584143b (pushed 2026-09-08): the 7 file-level issues above (individual replies inline), plus the worklist-completeness ask — added deterministic cues for the 8 rules you named without any (data-modeling/testing/performance/web-services) and, since I found the identical gap, for all 7 style rules from this PR too (15 cues total, across al-data-modeling-review.md, al-testing-review.md, al-performance-review.md, al-web-services-review.md, al-style-review.md). Ready for another look whenever you have time.

microsoft/knowledge/style/namespace-must-be-verified-from-source.md
JesperSchulz: The Markdown explanation is corrected, thanks. The companion bad fixture still says `// resolves in a local build, fails in VS Code` without showing stale/different symbols or configuration, so it continues to teach the inherent resolver discrepancy that the article now rejects. Please either make the guessed namespace fail against the same symbols in both environments, or explicitly model the differing symbol/configuration context. Then this thread is resolved.

microsoft/knowledge/data-modeling/check-post-line-batch-pattern.md
JesperSchulz: Thanks—the contextual caveat and the added Check Line fixture resolve part of the concern. The article still describes the responsibilities as “distinct, non-overlapping” and says Post Line writes “exactly one line to the ledger.” The standard General Journal implementation contradicts both claims: `Gen. Jnl.-Post Line.RunWithCheck` calls `Gen. Jnl.-Check Line.RunCheck`, and posting one journal line can create multiple G/L, VAT, balancing, rounding, and deferral entries.

Please describe the split as processing scope and reusable entry points instead: Check Line validates one journal line; Post Line processes/posts one journal line, may reuse validation, and may create multiple ledger entries; Post Batch orchestrates the batch.

microsoft/knowledge/data-modeling/dimension-management-wiring.md
JesperSchulz: Thanks—the prose now separates Default Dimensions from transactional Dimension Set IDs, which addresses the conceptual part of this thread. Two defects remain in the new good fixture:

1. `SaveDefaultDim(..., FieldNo("Global Dimension 1 Code"), ...)` passes table field ID 10, but that parameter is the shortcut-dimension number `1..8`; this example should pass `1`.
2. On customer change, `GetDefaultDimID` receives the document's existing Dimension Set ID with inherited table number `0`. That can preserve values inherited from the previous customer instead of recomputing the new customer's defaults. Please model this after a current document `CreateDim` implementation (normally recomputing source defaults without treating the existing document set as an anonymous inherited source), or explicitly define and test how user overrides are meant to win.

Once those fixture behaviors are corrected, I think this conversation can close.

JesperSchulz (changes requested):
Re-review of `584143b` against the exact head tree:

The frontmatter/structure validator, deterministic knowledge-index check (317 articles), and neutral review-fixture check (34 cases / 17 leaf domains) all pass, and the three reported checks are green. Changes are still required:

- The three unresolved discussions remain valid: the dimension fixture still passes field ID `10` where `SaveDefaultDim` requires shortcut number `1` and can retain the previous customer's dimensions; the posting article still claims non-overlapping responsibilities and exactly one ledger entry; the namespace bad fixture still teaches an inherent local-build/VS Code resolver discrepancy.
- `file-datatype-saas.good.al` uses the five-argument `UploadIntoStream` overload deprecated since runtime 7.0, declares `TempBlob` without using it, and does not follow the article's instruction to stage through `TempBlob`. Use the current overload and align the prose/fixture (or version-scope the example).
- `test-data-must-be-random-and-complete.md` and its routing cue classify a mandatory value as defective merely for being “underfilled”/“under-sized”. AL field lengths are maxima, not general minimums; limit this to missing values or an explicit semantic length requirement.
- `stored-derived-fields-must-not-be-exposed-directly.md` unnecessarily mandates expanding the API with source fields, while its own sample exposes only `Hours Used`, not the other formula input (`Budgeted Hours`), so the claimed consumer verification is impossible. Keep the API contract scoped to required fields; if verification is a requirement, expose all inputs.
- The branch currently conflicts with `main` in `al-breaking-changes-review.md`, `al-performance-review.md`, and `al-style-review.md`. Please rebase and preserve the deterministic cues when resolving them.

No other blocker found in this pass.

MichaelDieringer:
Thanks Jesper. Addressed the five open items from the 2026-09-15 re-review of `584143b`:

**The three unresolved discussions:**
- `dimension-management-wiring.good.al`: `SaveDefaultDim`'s third argument is the shortcut dimension *number* (1-8, indexing `General Ledger Setup`'s shortcut fields), not the AL field ID — the fixture passed `FieldNo("Global Dimension 1 Code")` (10). Fixed to the literal `1`, matching the `ValidateDimValueCode(1, ...)` call right above it. Also fixed `GetDefaultDimID`'s `InheritFromDimSetID` argument: the document fixture passed its own current `"Dimension Set ID"`, which inherits dimensions from whatever the document was previously linked to; changed to `0` so a customer change recomputes from scratch, matching the real pattern in `BankDepositHeader.Table.al`'s `CreateDim`. Verified both against `DimensionManagement.Codeunit.al` in the BCApps clone.
- `check-post-line-batch-pattern.md`: agreed "writes exactly one line to the ledger" overclaims — `Gen. Jnl.-Post Line` alone calls `InsertGLEntry` from a dozen call sites (balancing entry, VAT, currency rounding, deferrals) and can write several G/L Entries per journal line. Reworded to "posts exactly one journal line" and softened "distinct, non-overlapping responsibilities" to acknowledge real implementations aren't required to keep them non-overlapping.
- `namespace-must-be-verified-from-source.bad.al`: the prose fix from round 2 didn't reach the fixture — it still had `// resolves in a local build, fails in VS Code`. Replaced with a comment describing the actual cause (stale/cached symbols), matching the corrected prose.

**The two new findings:**
- `file-datatype-saas.good.al`: replaced the deprecated 5-argument `UploadIntoStream` overload with the current 2-argument one, and actually staged through `TempBlob` as the article's own Best Practice instructs (the declared `TempBlob` variable was unused before).
- `test-data-must-be-random-and-complete.md`: no longer treats a short-but-valid value as defective merely for being "underfilled" — AL field lengths are maxima, not minimums. Scoped to missing values or a scenario with an explicit length/format requirement (e.g. a truncation test), and updated the `al-testing-review.md` routing cue to match.

**Also fixed, found while verifying the above:**
- `stored-derived-fields-must-not-be-exposed-directly.md`/`.good.al`: the fixture exposed only one of the derived value's two inputs (`Hours Used`, not `Budgeted Hours`), so the claimed "expose the source field so the consumer can verify it" was actually unverifiable. Reframed source-field exposure as optional and all-or-nothing, and fixed the fixture to expose both.
- Found the same unlinked `See sample: `x.good.al`.` pattern already flagged on PR #161 repeated across 15 more articles in this PR — `Knowledge-Retrieval.ps1` needs the markdown-link form to associate a sample with its article. Fixed all 16.

Rebased onto current `main` (conflicts in `al-breaking-changes-review.md`, `al-data-modeling-review.md`, `al-performance-review.md`, `al-style-review.md` against merged #148/#153 — purely additive, both sides' worklist cues retained). All four local validators pass: frontmatter, knowledge-index, knowledge-retrieval (350 articles / 589 samples), review-fixtures (108 cases / 20 leaf domains).

JesperSchulz (changes requested):
Two merge-critical issues remain:

1. `api-page-key-fields-must-be-editable-on-insert.good.al` and `stored-derived-fields-must-not-be-exposed-directly.good.al` define writable API pages without `DelayedInsert = true`. That contradicts the existing `api-page-delayedinsert-true` rule and makes the canonical good samples teach code BCQuality itself flags. Add `DelayedInsert = true`, or make creation explicitly unsupported where appropriate.

2. `dimension-management-wiring.good.al` exits from `UpdateDimensionSetID` when the customer lookup fails. Clearing `Customer No.` therefore leaves the previous customer's shortcut dimension and dimension-set ID in place, contradicting the article's recompute-from-scratch requirement. Clear/recompute first and add the customer source only when it exists.

The previous routing, discovery, Email/File, testing, posting, namespace, and dimension-thread blockers otherwise appear resolved.

microsoft/knowledge/data-modeling/dimension-management-wiring.md
JesperSchulz: The original Default Dimension versus Dimension Set ID modeling issue and the two follow-up fixture defects are corrected. The latest review identified a separate customer-clearing cleanup defect and tracks it in the current change request, so this outdated thread can close.

microsoft/knowledge/data-modeling/check-post-line-batch-pattern.md
JesperSchulz: The article now describes processing scope and reusable entry points rather than non-overlapping responsibilities or one ledger entry per line. This concern is resolved—thanks for working through the distinction.

microsoft/knowledge/style/namespace-must-be-verified-from-source.md
JesperSchulz: The companion fixture now aligns with the corrected explanation and no longer teaches an inherent compiler-versus-language-server discrepancy. Resolved.

JesperSchulz:
Thanks for the substantial update. Most earlier concerns are now resolved, and I closed the three outdated review threads. The remaining review is narrowed to two concrete correctness fixes.

MichaelDieringer:
Thanks Jesper. Both fixed:

1. Added \`DelayedInsert = true\` to both writable API-page good samples (\`api-page-key-fields-must-be-editable-on-insert.good.al\`, \`stored-derived-fields-must-not-be-exposed-directly.good.al\`) — matches this repo's own \`api-page-delayedinsert-true\` rule.
2. \`dimension-management-wiring.good.al\`'s \`UpdateDimensionSetID\` no longer exits early on a failed \`Customer.Get\`. It now clears the shortcut field and calls \`GetDefaultDimID\` with an empty source list in that case too, so it correctly recomputes to an empty Dimension Set ID instead of leaving the previous customer's dimensions in place — the same staleness bug the \`InheritFromDimSetID = 0\` fix targeted last round, just reached via the other branch.

Validators pass clean.

JesperSchulz (approved):
The two remaining blockers are resolved: both writable API fixtures now set DelayedInsert, and the dimension sample clears/recomputes from scratch even when the customer is removed or missing. No new merge-critical issue found.

PR #157, opened 2026-09-04
18 more AL/BC patterns: data-modeling, testing, style, security, error-handling, ui, upgrade, web-services, appsource

MichaelDieringer:
## Summary

Second batch from CURABIS ApS — 18 more knowledge articles generalized from patterns observed across real AppSource/PTE Business Central development, follow-up to #156. Same format (6-key frontmatter, Description/Best Practice/Anti Pattern, sibling `.good.al`/`.bad.al` samples where a code contrast helps, no fenced code in the .md), targeted at existing Microsoft-owned domains rather than `/community/`.

- **data-modeling** (3): WorkDate must never be assigned in app code, Media/MediaSet over BLOB for pictures, the nine BC table-type design conventions (Master/Supplemental/Subsidiary/Ledger/Register/Journal/Document/Document History/Setup).
- **testing** (5): app-specific BCPT scenarios instead of generic samples, FEATURE/SCENARIO/GIVEN/WHEN/THEN tagging, `_UT` suffix for UI-layer test codeunits, GIVEN-block precondition completeness for posting/report scenarios, one WHEN per test (with flow-test and defect-then-fix exceptions).
- **style** (3): pages must not contain business logic, organize source by feature not object type, comments must not restate what the code already shows.
- **security** (1): every exposed object (API/web-service) must sit in a permission set.
- **error-handling** (2): log writes must survive rollback via an isolated session, defensive vs. offensive code should match actual blast radius.
- **ui** (1): the BC page-type design taxonomy (RoleCenter/Card/List/Worksheet/Document/etc.).
- **upgrade** (1): upgrade-tag guard logic must not nest deeply.
- **web-services** (1): give API pages least-privilege write access — a separate minimal page per writable field rather than widening a general-purpose one.
- **appsource** (1): app version bumps should be a deliberate major/minor decision, not left to automatic build/revision increments.

Checked against the current `microsoft/knowledge/` corpus before opening — four originally-drafted articles were dropped as duplicates or substantial overlaps of existing files (`xrec-is-a-before-image-only-in-some-triggers.md`, `use-library-codeunits-for-test-fixtures.md` + `asserterror-needs-expectederror-and-code.md`, `choose-telemetry-scope-by-audience.md`, `compose-permission-sets-with-included-sets.md`).

## Test plan


microsoft/knowledge/appsource/release-must-update-app-version.md
```diff
@@ -0,0 +1,37 @@
...
+
+At every release — a branch merged to `main`, a tagged release build, or an AppSource submission — the app's version is consciously updated, not left to the pipeline alone.
+
+| Version part | Owner | When |
+|---|---|---|
+| Major | Developer decision | Breaking change (schema, API, removed objects) |
+| Minor | Developer decision | Every release with new functionality |
+| Build / Revision | AL-Go pipeline | Automatic — never hand-edited |
+
+The version number is the only identity a deployed app has. Two customer environments running "the same" version with different code is an undiagnosable support case; an AppSource submission with an unchanged major.minor is a rejected submission. AL-Go increments build numbers on every CI run, which creates the illusion that versioning is handled — but major.minor is a human statement about compatibility, and no pipeline can make it.
```
JesperSchulz: AppSource requires the complete version to be greater than the previously submitted version; it does not require `major.minor` to change. An AL-Go-generated build increment can satisfy that requirement. “Build/revision are never hand-edited,” “feature branches never touch version,” and treating every merge to `main` as a release are workflow choices, not universal platform rules. Please reframe this around strict full-version ordering and scope any AL-Go branching policy explicitly.

microsoft/knowledge/data-modeling/pictures-must-use-media-not-blob.md
```diff
@@ -0,0 +1,44 @@
...
+---
+
+# Pictures must be stored in a Media/MediaSet field, not BLOB
+
+## Description
+
+`BLOB` is still a valid AL field type for arbitrary binary data, but it is
+not the right choice for storing pictures or images. The current
+recommendation is the `Media` field type for a single image, or
+`MediaSet` when a record needs several image variants (e.g. multiple
```
JesperSchulz: `MediaSet` stores multiple independent media objects; it is not an image-variant or multi-size abstraction. Thumbnails are generated per media object, and displaying multiple items requires custom handling. The recommendation should be `Media` for one image and `MediaSet` for multiple images/files, without promising automatic variant handling.

microsoft/knowledge/error-handling/log-writes-must-survive-rollback.good.al
```diff
@@ -0,0 +1,26 @@
...
+    begin
+        ErrorLogEntry.Init();
+        ErrorLogEntry."Call Duration (ms)" := CallDurationMs;
+        ErrorLogEntry."Error Message" :=
+            CopyStr(ErrorMessageText, 1, MaxStrLen(ErrorLogEntry."Error Message"));
+        ErrorLogEntry.Insert(true);
+        Commit();
+    end;
+
+    procedure SetParameters(Duration: Integer; ErrorText: Text)
```
JesperSchulz: This cannot pass values to a session started with `Session.StartSession`. The target codeunit is instantiated in another session with fresh globals, so calling `SetParameters` on a local instance does not populate `CallDurationMs` or `ErrorMessageText` there. `StartSession` can pass an optional record to a `TableNo` codeunit; the sample neither does that nor includes the caller. As written, the “good” logger inserts `0` and blank text. Please provide an end-to-end sample that transports the captured values across the session boundary.

microsoft/knowledge/security/exposed-objects-must-be-in-a-permission-set.md
```diff
@@ -0,0 +1,26 @@
...
+technologies: [al]
+countries: [w1]
+application-area: [all]
+---
+
+# Every exposed object must belong to a permission set
+
+## Description
+
+An object that is reachable from outside the app's own UI — an API page (`PageType = API`), a web-service-enabled page or query (`ServiceEnabled = true`), or a published API query — is only usable if it is also granted execute access through a permission set. When such an object is left out of every permission set, it becomes both unusable (no caller, human or service, can reach it) and invisible in review: nobody deliberately decided who may call it. Exposure without a matching grant is not a safe default; it is an endpoint nobody is governing.
```
JesperSchulz: `ServiceEnabled` is an attribute on codeunit methods, not a property that marks pages or queries as web-service enabled. Pages/queries are published through the Web Services configuration, while custom APIs use `PageType = API` / `QueryType = API`. Please correct the exposure signals; otherwise reviewers will search for invalid page/query syntax and miss actually published services.

microsoft/knowledge/style/pages-must-not-contain-business-logic.md
```diff
@@ -0,0 +1,31 @@
...
+technologies: [al]
+countries: [w1]
+application-area: [all]
+---
+
+# Keep business logic out of page objects
+
+## Description
+
+A page procedure that calculates a value and assigns it to a field, calls `Rec.Modify()` directly, or implements a business rule is an architecture violation even when it compiles. Pages are a presentation layer: they bind data to the UI and invoke actions. Calculations, validations, and record mutations belong in codeunits, where they can be tested, reused, and called consistently regardless of which page (or API, or batch job) triggers them. When logic lives on a page, it only applies when a user opens that specific page — the same business rule silently doesn't run through any other entry point.
```
JesperSchulz: The entry-point-independence principle is sound, but “calculations, validations, and record mutations belong in codeunits” is too absolute. Reusable data invariants commonly belong in table fields/triggers/procedures, and presentation-specific calculations or state belong on pages. Please scope the rule to cross-entry-point business operations and name tables as a valid owner; the current wording would make agents flag conventional Base Application code and encourage moving table invariants into codeunits.

microsoft/knowledge/testing/test-feature-scenario-tags.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Best Practice
+
+Put `[FEATURE]` as a comment before the codeunit's opening brace, naming the domain rather than the object. Put `[SCENARIO]` immediately above each `[Test]` attribute, describing the scenario in business language that complements — not duplicates — the procedure name. Inside the body, mark the precondition setup as `[GIVEN]`, the single action under test as `[WHEN]`, and the assertions as `[THEN]`. The procedure name stays the machine-readable identity shown in test-runner output; the `[SCENARIO]` comment stays the human-readable one. Neither replaces the other.
```
JesperSchulz: This exact placement contradicts the current BCApps test corpus: Microsoft tests commonly put `[FEATURE]` and `[SCENARIO ...]` inside each test procedure after `begin`, followed by GIVEN/WHEN/THEN. The useful rule is clear scenario structure, not requiring FEATURE before the object brace and SCENARIO before `[Test]`. Please accept the canonical in-procedure form so this does not flag existing Microsoft tests as defective.

microsoft/knowledge/testing/ui-test-codeunit-naming.md
```diff
@@ -0,0 +1,26 @@
...
+
+# Suffix UI-layer test codeunits with _UT and never mix layers in one codeunit
+
+## Description
+
+A test codeunit that drives pages through `TestPage` — opening pages, reading FactBox parts, triggering field `OnValidate` through the page — is testing a different layer than a codeunit that calls business-logic procedures directly. Readers need to know which layer a given test exercises without opening it, and a single codeunit that mixes both kinds of test hides that distinction: a failure could mean the logic broke, the page broke, or both.
+
+## Best Practice
+
+Give any test codeunit that uses `TestPage` a `_UT` (Unit Test — UI layer) suffix, and keep it free of tests that call codeunit/table procedures directly. Keep the corresponding logic-only codeunit unsuffixed. Allocate the two codeunits adjacent object IDs so their relationship is visible in the object list.
```
JesperSchulz: `_UT` is not a BC convention meaning “UI layer.” BCApps uses UT for unit tests generally, has many TestPage-driven tests in codeunits without this suffix, and sometimes uses `_UT` on individual method names inside broader suites. Adjacent object IDs are also not a semantic pairing mechanism. This is a CURABIS naming convention, so it should not be a universal Microsoft-layer anti-pattern unless it is narrowed to projects that explicitly adopt that convention.

microsoft/knowledge/ui/page-design-must-match-bc-page-type-conventions.md
```diff
@@ -0,0 +1,76 @@
...
+link back to.
+
+## Best Practice
+
+Match the page's design to its type:
+
+- **RoleCenter** — tailored home page for a role; named role + `Role
+  Center`; links to List pages, shows Cues/Activities.
+- **Card** — view/edit one record; named table + `Card`; FastTabs only,
+  first FastTab named `General`. Requires a single-field primary key — a
```
JesperSchulz: Card pages do not require a single-field primary key. Current Base Application counterexamples include `Customer Bank Account Card`, `Vendor Bank Account Card`, and `Ship-to Address`, all Card pages over composite-key tables. The later blanket `UsageCategory` warning is also unsafe because non-searchable supporting pages intentionally omit it. Please turn these into contextual heuristics rather than mandatory structural constraints.

microsoft/knowledge/web-services/api-page-least-privilege-write-access.md
```diff
@@ -0,0 +1,26 @@
...
+technologies: [al]
+countries: [w1]
+application-area: [all]
+---
+
+# Give API pages least-privilege write access
+
+## Description
+
+A general-purpose API page that exposes many fields should not be widened to allow writes on one additional field. A `PageType = API` page consumed by an external integration, an automation agent, or a partner system carries the same risk regardless of caller: a write-enabled page with no per-field restriction is a wide-open surface. The most common real-world shape of this problem is not a page that started narrow and got widened — it is a page that was never restricted at all: with `InsertAllowed`/`ModifyAllowed`/`DeleteAllowed` left at their defaults and no `Editable = false` on any field, every field on the source table — including identity fields and financially significant ones — is fully writable, with nothing marking that as deliberate.
```
JesperSchulz: Only fields included on the API page are exposed/writable; default editability does not make every field on the source table writable. More importantly, least privilege must cover operations as well as fields. The proposed good fixture leaves `InsertAllowed` and `DeleteAllowed` at their `true` defaults, so an endpoint supposedly created only to update contact email can also create and delete Vendor records. Please correct the claim and set the unused operations to `false` in the good sample.

microsoft/knowledge/testing/given-blocks-must-cover-full-precondition-chain.good.al
```diff
@@ -0,0 +1,15 @@
+[Test]
+procedure PostSalesOrder_CreatesInvoice()
+var
+    Customer: Record Customer;
+    SalesHeader: Record "Sales Header";
+begin
+    // [GIVEN] a customer with a full posting-group chain and VAT setup
+    LibrarySales.CreateCustomerWithPostingSetup(Customer);
```
JesperSchulz: These are presented as `Library - Sales` APIs, but `CreateCustomerWithPostingSetup`, `CreateSalesOrderForCustomer`, and `PostSalesOrder` are not procedures on the current library codeunit. The real APIs include shapes such as `CreateCustomer`, `CreateSalesOrderForCustomerNo`, and `PostSalesDocument`. Please use real library calls or define these as local wrappers in a complete fixture; future agents should not learn nonexistent APIs from the good example.

JesperSchulz (changes requested):
There are useful principles in this batch, but it is not ready for the Microsoft layer yet. I verified several concrete platform/source contradictions inline.

There are also two cross-cutting blockers:

1. None of the 18 articles is wired into the corresponding `microsoft/skills/review/al-*-review.md` worklist. The current contracts say their targeted checks cover every current article, but no changed skill names any new slug and several domains' relevance gates would return `not-applicable` before considering these topics. Please add explicit signals/targeted checks and coordinate with #155's coverage contract so agents can reliably consume the knowledge.
2. Several entries are CURABIS house conventions presented as universal Microsoft requirements: feature-vs-object folder layout, exactly one WHEN, `_UT` meaning UI-layer tests, adjacent object IDs, and a hard two-level nesting limit. These can be useful team guidance, but they need authoritative BC-wide grounding and careful scoping before an agent may emit gating findings from the Microsoft layer; otherwise they belong in a custom/community layer.

The structural validators pass (with two keyword-count warnings), but they do not compile the AL fixtures or validate the claims. The CLA is still pending; because this is generalized from CURABIS work, the contributor must use the applicable employer/company declaration and confirm authorization to contribute it.

microsoft/knowledge/appsource/release-must-update-app-version.md
MichaelDieringer: You're right, and thanks for catching it precisely. On reflection this is a CURABIS house convention leaking through as a stated Microsoft requirement. We always bump minor deliberately per feature release and never hand-edit build/revision, and that habit got written up as if it were an AppSource platform rule rather than our own release discipline. 
Will reframe this around strict full-version ordering (the actual platform requirement) and either scope the major/minor-discipline guidance explicitly as a team convention, or move it to custom/community. Appreciate the rigor, but it's sharpening our own contribution pipeline too.

microsoft/knowledge/data-modeling/pictures-must-use-media-not-blob.md
MichaelDieringer: Yes, MediaSet doesn't do anything variant/thumbnail-related on its own, that was an unfounded assumption in the original wording. Corrected framing: Media for a single picture, MediaSet for several independent pictures/files, no automatic derived images either way.

-----------------------------------

For the record, if anyone reading this needs an actual document-preview pattern: we've built exactly this for a customer. Keep the source file in one Media field, generate a PNG of page 1 via the platform's Codeunit "PDF Document" (Load + ConvertToImage), and ImportStream it into a second Media field (Preview), regenerated on import with a manual "Regenerate" action as a fallback. Happy to turn that into a proper follow-up knowledge article with a verified .good.al sample if that's useful for the corpus.

microsoft/knowledge/error-handling/log-writes-must-survive-rollback.good.al
MichaelDieringer: Confirmed against the Session data type reference, you're completely right, and it's a more fundamental miss than I initially registered: StartSession's only channel for transporting data into the new session is the optional Record parameter, delivered to the target codeunit's OnRun trigger (which requires TableNo). Calling SetParameters on a local instance before starting the session doesn't do anything useful. Each session gets its own fresh instantiation, no shared state. As written the sample would silently insert 0/blank, exactly as you say.

We'll rework this as: pass the duration/error text via a Record parameter to a TableNo-based codeunit, and have the log write happen and commit inside the started session itself. That's actually the correct shape for "survives caller rollback" anyway, since the point isn't to get data back to the caller, it's to make sure the write isn't undone by the caller's transaction.

More broadly: this is a good concrete case for your blocker #3. The fixture was never actually run against a BC container, only structurally validated, so a plausible-looking-but-broken sample sailed through. We're adding an execution/compile step for .good.al/.bad.al fixtures to our own pre-submission process going forward, not just format validation.

microsoft/knowledge/style/pages-must-not-contain-business-logic.md
MichaelDieringer: You're right, and on closer inspection this is actually the same failure as the app-version article: our own source rule explicitly says "this standard is more rigorous than general Business Central guidance". A deliberate CURABIS house policy, but that caveat got dropped when it was generalized for this submission, leaving only the absolute claim.
Table-owned invariants and presentation-specific calculations are legitimate, as you say; the part we actually want to keep is narrower. Don't perform persisted write mutations (Rec.Modify() etc.) or cross-entry-point business rules from a page.

Also flagging something we found ourselves while checking: the "BCApps Reference" section only verified one example (Performance Toolkit setup pages) and then claimed "Microsoft applies this uniformly across thousands of pages in BCApps". An overgeneralization from a single sample that we didn't check against counterexamples. Will rescope both the rule and that reference before resubmitting.

microsoft/knowledge/testing/test-feature-scenario-tags.md
MichaelDieringer: One data point worth adding, just for alignment sake: Microsoft's own docs on this topic (https://learn.microsoft.com/dynamics365/business-central/dev-itpro/developer/devenv-extension-advanced-example-test#describe-your-tests) explicitly describe [FEATURE] as settable once at codeunit level, inherited by all tests in it, which matches our placement and also matches how we structure test codeunits (one per feature, tests derived from there).

[SCENARIO] placement isn't as clearly pinned down in that doc, so if the actual BCApps corpus consistently puts it inside the procedure body after begin, we'll adapt ours to match rather than keep a divergent form. 

I'm happy to align with whatever's canonical in the current codebase rather than introduce a second convention.

microsoft/knowledge/testing/ui-test-codeunit-naming.md
MichaelDieringer: Fair, _UT and the adjacent-ID pairing are both our own conventions, not BCApps ones; neither was checked against actual BCApps naming practice before this went in, and framing them as a universal anti-pattern was the mistake, not the conventions themselves. 
We'll narrow this to explicitly scoped team guidance rather than a Microsoft-layer rule.

microsoft/knowledge/ui/page-design-must-match-bc-page-type-conventions.md
MichaelDieringer: You're right, and this one's worse than a scoping miss. The source file explicitly claims "cross-checked against the current Base Application page structure... the taxonomy still holds," but that check evidently didn't catch Customer/Vendor Bank Account Card or Ship-to Address. 
We'll turn the primary-key claim into a contextual heuristic rather than a mandatory constraint, per your suggestion. 

There's also a real pattern our table-type companion rule misses entirely: a subsidiary table that supplements a master record with its own identity (parent key + own code). Ship-to Address being the clearest example, commonly gets its own Card page for direct editing, not the Worksheet/Tabular shape we'd documented for subsidiary tables generally. Will fix both together.

microsoft/knowledge/web-services/api-page-least-privilege-write-access.md
MichaelDieringer: You're right on both counts. On the "every field on the source table" claim. That's a generalization slip on our part: the original example was scoped to a page where ~90 fields were all actually placed on it, which is accurate, but that scoping got lost when we generalized it and it now reads as if any unrestricted API page exposes the whole table regardless of what's on it. Will correct.

The InsertAllowed/DeleteAllowed gap is the sharper one. That's in our own "good" example from the start, not introduced in this submission. For what it's worth, our actual MCP setup restricts the write operation at the MCP tool configuration layer rather than relying on the page alone, but the fixture itself should reflect least-privilege properly regardless. We'll set the unused operations to false as you suggested. Appreciate you catching it.

microsoft/knowledge/testing/given-blocks-must-cover-full-precondition-chain.good.al
MichaelDieringer: Confirmed against the current LibrarySales codeunit, CreateCustomerWithPostingSetup and CreateSalesOrderForCustomer don't exist as you say. One small correction to your note: PostSalesOrder does exist as a name, but with a different signature (it creates and posts a line in one call, not just posts an existing header). 
Calling it the way this fixture did would still fail to compile, just not because the name is missing. Either way, you're right that this fixture is broken.
This fixture was adapted from one of our own existing tests for the public submission, but the generalization step lost the exact library API names and wasn't re-verified against the current codeunit before we submitted it. Will rewrite against the real API or as explicit local wrappers.

MichaelDieringer:

MichaelDieringer:

JesperSchulz (commented):
Follow-up after `81845a5`: the version ordering, API restrictions, FEATURE/SCENARIO placement, `_UT`/object-ID claims, composite-key Card guidance, and several earlier fixture issues are improved. The remaining correctness issues are inline below.

The deterministic testing worklist is also incomplete: only a subset of the 18 new rules has an explicit cue, and several cues recognize compliant code rather than the anti-pattern. Please ensure every new rule can be selected from the changed-code signals that indicate a possible violation. The invented API comment below applies to all new good fixtures; they should be compiled against the declared dependencies rather than treated as pseudocode.

microsoft/knowledge/error-handling/log-writes-must-survive-rollback.good.al
```diff
@@ -0,0 +1,45 @@
...
+}
+
+codeunit 50100 "Sample Error Log Writer"
+{
+    // TableNo makes OnRun receive the Record that Session.StartSession
+    // passes to the new session. This is the only data channel into that
+    // session — there is no shared memory with the caller's instance.
+    TableNo = "Sample Error Log Buffer";
+
+    trigger OnRun(var Rec: Record "Sample Error Log Buffer")
```
JesperSchulz: This trigger declaration is invalid AL. A codeunit `OnRun` trigger is declared as `trigger OnRun()`; when `TableNo` is set, `Rec` is available implicitly and is not a formal parameter. Please correct the fixture and compile it—the current “good” example teaches syntax that will fail immediately.

microsoft/knowledge/security/exposed-objects-must-be-in-a-permission-set.md
```diff
@@ -0,0 +1,26 @@
...
+technologies: [al]
+countries: [w1]
+application-area: [all]
+---
+
+# Every exposed object must belong to a permission set
+
+## Description
+
+An object that is reachable from outside the app's own UI is only usable if it is also granted execute access through a permission set. Three distinct mechanisms make an object reachable this way, and each needs to be checked on its own terms: a page or query published through the **Web Services** configuration page; a custom REST endpoint declared with `PageType = API` / `QueryType = API`; or an individual codeunit method exposed with the `[ServiceEnabled]` attribute (a method-level attribute — it does not apply to pages or queries as a property). When such an object is left out of every permission set, it becomes both unusable (no caller, human or service, can reach it) and invisible in review: nobody deliberately decided who may call it. Exposure without a matching grant is not a safe default; it is an endpoint nobody is governing.
```
JesperSchulz: This still describes `[ServiceEnabled]` too broadly. It is used on page procedures for API bound actions; public procedures on a codeunit published as a web service provide unbound operations without adding this attribute to each method. The permission guidance must also cover `codeunit "..." = X` for a published codeunit, not only page/query execute entries. Please distinguish the exposure mechanisms and their corresponding permission targets.

microsoft/knowledge/data-modeling/code-must-not-change-workdate.md
```diff
@@ -0,0 +1,40 @@
...
+application-area: [all]
+---
+
+# Application code must not change the WorkDate
+
+## Description
+
+The work date is a per-user session setting the user controls from the
+client (the date shown in the top-right corner, used to default posting
+dates and date filters). Application code must never call the `WorkDate`
```
JesperSchulz: “Must never” is too broad. Application code legitimately writes `WorkDate` when implementing the user's explicit work-date/user-settings action, and tests may deliberately control it as test context. The defect is unrelated business logic silently mutating session state. Please scope the rule to that behavior instead of prohibiting the setter itself in every context.

microsoft/knowledge/testing/bcpt-scenarios-must-be-app-specific.md
```diff
@@ -0,0 +1,26 @@
...
+
+# Include app-specific scenarios in a PerformanceTest app's BCPT suite
+
+## Description
+
+A PerformanceTest app that ships with only the generic Microsoft BCPT samples (creating sales orders, purchase orders, posting item journals) measures Business Central's own baseline performance, not the extension it was built to test. Those samples are starting points, not coverage. Without a scenario that exercises the extension's own business flow — its own codeunits, its own FlowFields, its own page rendering — a performance regression introduced by the extension has no test that would ever detect it.
+
+## Best Practice
+
+For every major business flow the extension adds, create a matching `BCPT*` scenario codeunit: `SingleInstance = true`, implementing `"BCPT Test Param. Provider"`, wrapping the operation under test in `BCPTTestContext.StartScenario()` / `EndScenario()`, and building its own test data in a local `InitTest()` procedure rather than depending on hardcoded records. Give each distinct step its own named scenario so a regression in one step doesn't hide inside a coarser measurement.
```
JesperSchulz: These are not universal BCPT requirements. Microsoft's current `BCPT Create Customer` sample is neither `SingleInstance` nor wrapped in `StartScenario`/`EndScenario`. Requiring every scenario to implement all of these details would make agents reject valid BCPT code. Keep the app-specific coverage principle, but describe the interface, lifecycle calls, and scenario granularity as context-dependent patterns backed by the BCPT version/API being targeted.

microsoft/knowledge/testing/test-feature-scenario-tags.good.al
```diff
@@ -0,0 +1,26 @@
...
+    [Test]
+    procedure GetPrice_CustomerPrice_ReturnsUnitPrice()
+    var
+        Customer: Record Customer;
+        Item: Record Item;
+        UnitPrice, LineDiscPct: Decimal;
+    begin
+        // [SCENARIO] Customer with a specific price list line gets that unit price
+        // [GIVEN] a customer with a price list line at 100 LCY
+        LibrarySales.CreateCustomerWithPrice(Customer, Item, '', 100);
```
JesperSchulz: `Library - Sales` does not expose `CreateCustomerWithPrice`, and the other new fixtures still contain similar invented APIs such as `CreateCustomerWithTierPrice` and `"Item Price Tier Buffer"`. A “good” fixture must compile against real BCApps test-library APIs. Please replace these with valid setup code and compile all fixtures, rather than using plausible-looking pseudocode.

microsoft/knowledge/ui/page-design-must-match-bc-page-type-conventions.md
```diff
@@ -0,0 +1,83 @@
...
+## Anti Pattern
+
+A page that mixes conventions from two types — for example, a "List"
+page with no `CardPageID` even though a Card page exists for the same
+table — signals a design step was skipped, not a stylistic choice. A
+Card page over a composite-key table is not automatically this anti
+pattern; check whether the table supplements a master record first. Also watch
+for: a Worksheet or List page showing primary-key fields it shouldn't (or
+hiding them when it should show them), and a page with no
+`UsageCategory` set, which makes it invisible to Tell Me search even
```
JesperSchulz: An omitted `UsageCategory` is not inherently a page-design defect. Supporting pages, subpages, dialogs, and pages intended only to be reached through another workflow often should not be discoverable in Tell Me. Please flag this only when the page is intended as a searchable entry point; otherwise agents will add navigation exposure that the design did not request.

microsoft/knowledge/error-handling/defensive-vs-offensive-code-must-match-blast-radius.md
```diff
@@ -0,0 +1,26 @@
...
+
+See sample: `defensive-vs-offensive-code-must-match-blast-radius.good.al`.
+
+## Anti Pattern
+
+Guarding two fields the same way purely out of habit, without analyzing what each one feeds. A low-blast-radius field, such as a VAT registration number shown only on a printed document, and a high-blast-radius field, such as the VAT posting group that determines VAT actually applied to a posted transaction, are both wrapped in the same `if Header.Get(...) then ... else` pattern with a blank/zero fallback — leaving the posting-critical field free to post with a silently wrong value.
```
JesperSchulz: A VAT registration number on a document is not a safe generic example of low blast radius: it can be legally relevant, validated, and involved in external VAT services or mandated document output. Please use a genuinely cosmetic field for the low-risk side, or explicitly qualify the narrow display-only scenario so agents do not learn to silently default tax identity data.

microsoft/knowledge/style/source-organized-by-feature-not-object-type.md
```diff
@@ -0,0 +1,36 @@
...
+    src/
+    ├── NoSeries/
+    ├── Sales/
+    │   ├── Invoice/
+    │   └── Order/
+    └── Common/
+
+Each feature folder holds every object type it needs; shared code has one dedicated home.
+
+## Anti Pattern
```
JesperSchulz, line 29: The description correctly calls this a deliberate team convention, but this section still labels the object-type layout itself an anti-pattern. Those positions conflict. The anti-pattern should be inconsistency within a repository or violating its documented convention—not choosing one of two reasonable organization strategies.

microsoft/knowledge/data-modeling/pictures-must-use-media-not-blob.md
```diff
@@ -0,0 +1,49 @@
...
+explicitly and stored in its own field, regardless of which type holds the
+source image.
+
+`BLOB` remains the correct choice for genuinely arbitrary binary payloads
+that are not images and don't benefit from the media pipeline (e.g. a raw
+file attachment blob unrelated to picture rendering).
+
+## Best Practice
+
+Use `Media` (or `MediaSet` for multiple image variants) for any field that
```
JesperSchulz: This says `MediaSet` is for “image variants” immediately after correctly explaining that it stores independent media objects and does not generate variants. Please use “multiple independent images” here as well; derived sizes/variants need to be created and stored explicitly.

microsoft/knowledge/error-handling/log-writes-must-survive-rollback.good.al
MichaelDieringer: Fixed in 30abf07 — corrected to `trigger OnRun()`. `Rec` is available implicitly once `TableNo` is set on the codeunit; it's not a formal trigger parameter.

microsoft/knowledge/security/exposed-objects-must-be-in-a-permission-set.md
MichaelDieringer: Fixed in 30abf07 — distinguished the three exposure mechanisms and their actual permission targets: a published/API page or query needs `page "..." = X` / `query "..." = X`; a codeunit published through Web Services exposes every public procedure automatically with no per-method attribute, so the target is `codeunit "..." = X`; `[ServiceEnabled]` is specifically a page-procedure attribute for an OData v4 bound action, not an object-level property on pages/queries/codeunits, and it's still governed by that page's own `page "..." = X` entry rather than a permission target of its own.

microsoft/knowledge/data-modeling/code-must-not-change-workdate.md
MichaelDieringer: Fixed in 30abf07 — scoped from an absolute "never" to "not as a side effect of unrelated logic." Verified two real BCApps counter-examples: `CreateDemoEDocsBE.Codeunit.al` saves the work date, sets a specific one to backdate demo data, and restores it afterward (`WorkDate(SampleInvoiceDate)` / `WorkDate(SavedWorkDate)`), and `SustainabilityPostingTest.Codeunit.al` deliberately sets `WorkDate` to control the date context a test runs under. Both are the code's actual purpose, not an incidental side effect — the anti-pattern is now scoped to unrelated business logic silently mutating that state, not the setter form itself.

microsoft/knowledge/testing/bcpt-scenarios-must-be-app-specific.md
MichaelDieringer: Fixed in 30abf07 — reframed `SingleInstance` and `StartScenario`/`EndScenario` from mandatory requirements to context-dependent patterns tied to the BCPT version/API being targeted, matching that Microsoft's own `BCPT Create Customer` sample uses neither. The app-specific coverage principle stays; the implementation-detail requirements don't.

microsoft/knowledge/testing/test-feature-scenario-tags.good.al
MichaelDieringer: Fixed in 30abf07 — replaced the invented calls with a real, verified price-list-line test using `Library - Sales`/`Library - Inventory`/`Library - Price Calculation`. Verified each call directly against BCApps source: `LibraryPriceCalculation.CreatePriceHeader`/`CreateSalesPriceLine`, `LibrarySales.CreateCustomer`/`CreateSalesDocumentWithItem`, and `LibraryInventory.CreateItem` all exist with exactly the signatures the fixture calls. While sweeping the rest of this PR's fixtures for the same pattern, also fixed two more that weren't flagged in your review but had the identical problem: `given-blocks-must-cover-full-precondition-chain.bad.al` (wrong arity, undeclared variable) and `ui-test-codeunit-naming.good.al`/`.bad.al` (same fake `"Item Price Mgt."`/TestPage "Item Price", replaced with real `Library - Sales` calls and the real Customer Card TestPage).

microsoft/knowledge/ui/page-design-must-match-bc-page-type-conventions.md
MichaelDieringer: Fixed in 30abf07 — scoped the missing-`UsageCategory` anti-pattern to pages intended as searchable entry points. Supporting pages, subpages, dialogs, and pages only reached through another workflow are explicitly excluded, so agents won't add navigation exposure the design didn't ask for.

microsoft/knowledge/error-handling/defensive-vs-offensive-code-must-match-blast-radius.md
MichaelDieringer: Fixed in 30abf07 — replaced the VAT registration number example with a genuinely cosmetic field (customer home page URL) for the low-risk side, so the article no longer teaches agents to silently default tax-identity data.

microsoft/knowledge/style/source-organized-by-feature-not-object-type.md
MichaelDieringer: Fixed in 30abf07 — reframed the anti-pattern as inconsistency within a repo or violating its own documented convention, not choosing the object-type layout itself. The two sections no longer contradict each other.

microsoft/knowledge/data-modeling/pictures-must-use-media-not-blob.md
MichaelDieringer: Fixed in 30abf07 — removed the leftover "image variants" wording; the article now consistently says "multiple independent images" throughout, matching the already-corrected MediaSet description.

MichaelDieringer:
@JesperSchulz Addressed this round in 30abf07 (pushed 2026-09-08): the 9 file-level issues above (individual replies inline), plus two more invented-API fixtures found in the same sweep that weren't in your review (`given-blocks-must-cover-full-precondition-chain.bad.al`, `ui-test-codeunit-naming.good.al`/`.bad.al` — noted inline on the price-fixture thread). Also added worklist cues for the 12 of 18 new rules that had none (al-appsource-review.md, al-data-modeling-review.md, al-error-handling-review.md, al-security-review.md, al-style-review.md ×3, al-testing-review.md ×2, al-ui-review.md, al-upgrade-review.md, al-web-services-review.md), and fixed test-feature-scenario-tags' own cue, which only matched the compliant tagged shape instead of the anti-pattern. Ready for another look whenever you have time.

microsoft/knowledge/web-services/api-page-least-privilege-write-access.good.al
```diff
@@ -0,0 +1,23 @@
+page 50102 "Vendor Contact Info API"
+{
+    PageType = API;
```
JesperSchulz, line 3: The API-page examples are still missing mandatory `EntityName` and `EntitySetName` properties. Please add distinct singular and plural camelCase values to both the good and bad fixtures and compile them against the normal Base Application symbols. This matters because the files are presented as complete API page objects, not abbreviated fragments.

microsoft/knowledge/style/pages-must-not-contain-business-logic.good.al
```diff
@@ -0,0 +1,31 @@
+codeunit 50100 "Sales Line Management"
+{
+    procedure RecalculateLine(var SalesLine: Record "Sales Line")
+    begin
+        SalesLine."Total Amount" := SalesLine.Quantity * SalesLine."Unit Price";
```
JesperSchulz, line 5: `Sales Line` has no field named `"Total Amount"` in the current Base Application, so both this good fixture and the corresponding bad fixture do not compile. Please use an actual field appropriate to the example—such as `Amount`, `"Line Amount"`, or a declared custom field/table—while preserving the intended page-versus-domain-logic lesson.

microsoft/knowledge/testing/test-feature-scenario-tags.good.al
```diff
@@ -0,0 +1,37 @@
...
+        PriceListHeader: Record "Price List Header";
+        PriceListLine: Record "Price List Line";
+        SalesHeader: Record "Sales Header";
+        SalesLine: Record "Sales Line";
+    begin
+        // [SCENARIO] Customer with a specific price list line gets that unit price
+        // [GIVEN] a customer and an item with a customer-specific sales price list line
+        LibrarySales.CreateCustomer(Customer);
+        LibraryInventory.CreateItem(Item);
+        LibraryPriceCalculation.CreatePriceHeader(
```
JesperSchulz, line 26: The replacement APIs now exist, thanks. One behavioral step is still missing: `CreatePriceHeader` creates a Draft price list, and Draft lists are not used for price calculation. Please validate the header to `"Price Status"::Active` and persist it before creating the sales line. The same setup is duplicated in `test-one-when-per-test.good.al` and needs the same correction; otherwise the “good” tests do not exercise the price list they assert against.

microsoft/knowledge/security/exposed-objects-must-be-in-a-permission-set.md
```diff
@@ -0,0 +1,32 @@
...
+---
+
+# Every exposed object must belong to a permission set
+
+## Description
+
+An object that is reachable from outside the app's own UI is only usable if it is also granted execute access through a permission set. Three distinct mechanisms make an object reachable this way, each with its own permission target:
+
+- A page or query published through the **Web Services** configuration page, or a custom REST endpoint declared with `PageType = API` / `QueryType = API` — both need a `page "..." = X` / `query "..." = X` entry for that object.
+- A codeunit published through **Web Services** exposes *every* public procedure on it as an OData/SOAP operation automatically — there is no per-method attribute to add. The permission target is the codeunit itself: `codeunit "..." = X`.
```
JesperSchulz, line 17: The execute-permission correction is good, but a published codeunit is not an OData service. Current Business Central supports Codeunit and Page for SOAP publication, while Page and Query are the OData object types. Please describe published codeunits as SOAP services and recommend API pages/queries for new REST integrations. This distinction is especially important now that SOAP is deprecated.

JesperSchulz (commented):
Thank you for another substantial round of corrections. Most of the previous findings are now resolved, and this PR is much closer.

During verification of the updated examples, we found a small number of correctness issues that the structural validators cannot detect—for example, AL that does not compile, test setup that never activates the data it intends to exercise, or guidance that differs from the current platform behavior. We recognize that another round is inconvenient, and we are not trying to prolong the review. These articles will be consumed as authoritative instructions by automated agents, so an incorrect “good” fixture or overly broad rule can be repeated across many future changes. We have therefore limited this round to the material issues below and omitted optional editorial cleanup.

microsoft/skills/review/al-error-handling-review.md
```diff
@@ -47,6 +47,8 @@ A file enters the candidate worklist when its `keywords` intersect the extracted
 The following targeted checks cover every current `error-handling` article:
 
 - `[ErrorBehavior(ErrorBehavior::Collect)]`, `ErrorInfo.Collectible`, `HasCollectedErrors`, `GetCollectedErrors`, or `ClearCollectedErrors` is added or changed, especially when errors are collected without later surfacing/clearing them — `collect-validation-errors-with-errorbehavior`.
+- New or changed code calls `Session.StartSession` from within error/duration logging around a web-service call, background job, or other operation expected to fail — `log-writes-must-survive-rollback`.
```
JesperSchulz, line 50: This cue still selects the article from `Session.StartSession`, which is the compliant remedy, while the bad fixture contains no such call. Please cue on the actual risk: inserting a failure/error log in the current transaction around a failed `TryFunction`/`GetLastError*` path and then raising, propagating, or rethrowing the error. `StartSession` can then be treated as a compliant discriminator rather than the entry condition.

microsoft/knowledge/ui/page-design-must-match-bc-page-type-conventions.md
```diff
@@ -0,0 +1,85 @@
...
+application-area: [all]
+---
+
+# Pages must match one of Business Central's page-type conventions
+
+## Description
+
+Business Central's page types — RoleCenter, Card, List, CardPart,
+ListPart, Worksheet, Document, ListPlus, plus the system dialog types
+(Navigate, ConfirmationDialog, StandardDialog, HeadlinePart, API) — each
```
JesperSchulz, line 16: Because this is marked `[all]` and reads as a page-type catalogue, please use the exact current enum name `NavigatePage`, not `Navigate`, and either include current types such as `PromptDialog`, `ConfigurationDialog`, and `UserControlHost` or state clearly that this is only a selected list of conventional types. Otherwise an agent can reject a valid current `PageType` or suggest an invalid one.

microsoft/knowledge/appsource/release-must-update-app-version.md
```diff
@@ -0,0 +1,37 @@
...
+
+At every release — a branch merged to `main`, a tagged release build, or an AppSource submission — the app's version is consciously updated, not left to the pipeline alone.
+
+| Version part | Owner | When |
+|---|---|---|
+| Major | Developer decision | Breaking change (schema, API, removed objects) |
+| Minor | Developer decision | Every release with new functionality |
+| Build / Revision | AL-Go pipeline | Automatic — never hand-edited |
+
+The version number is the only identity a deployed app has. Two customer environments running "the same" version with different code is an undiagnosable support case. AppSource's actual requirement is strict full-version ordering — the complete version must be greater than the previously submitted version — which an AL-Go-generated build/revision increment can satisfy on its own; AppSource does not require major.minor itself to change. Treating major.minor as a deliberate, human-decided compatibility signal is still valuable practice — it is a statement about what changed that no pipeline can make on its own — just not a platform-enforced requirement.
```
JesperSchulz, line 22: `id` is the app's stable identity; `version` identifies a release of that app. Calling the version “the only identity” is factually wrong and can misteach manifest semantics. Please reword this to say that the version distinguishes deployed code states/releases, while retaining the valid warning about different code sharing one version.

microsoft/knowledge/error-handling/defensive-vs-offensive-code-must-match-blast-radius.good.al
```diff
@@ -0,0 +1,10 @@
+// Low blast radius: guard, with an explicit chosen fallback.
+if Customer.Get(SalesHeader."Sell-to Customer No.") then
+    CustomerHomePage := Customer."Home Page";
+// Blank is an acceptable, deliberately-considered default here - the field
```
JesperSchulz, line 4: This is not an explicit fallback: when `Customer.Get(...)` is false, no assignment occurs, so `CustomerHomePage` retains its prior/default value. That contradicts the article's own requirement for an explicit, deliberately chosen fallback. Add an `else` assignment (or clear/initialize the value explicitly before the lookup) so the good fixture demonstrates the stated rule.

microsoft/knowledge/testing/bcpt-scenarios-must-be-app-specific.good.al
```diff
@@ -0,0 +1,43 @@
...
+
+    local procedure InitTest()
+    begin
+        // Set up any required configuration
+    end;
+
+    local procedure CreateServiceRequest(var BCPTTestContext: Codeunit "BCPT Test Context")
+    begin
+        BCPTTestContext.StartScenario('Create Service Request Header');
+        // ... create the service request
```
JesperSchulz, line 26: The “good” fixture measures no app-specific work: `InitTest` is empty and both measured sections contain only comments. It therefore has the same behavioral defect the article warns about despite its app-specific names. Please include a real, self-contained sample operation/data path in the measured sections so this fixture actually exercises extension logic.

microsoft/knowledge/upgrade/upgrade-tag-logic-must-not-nest-deeply.good.al
```diff
@@ -0,0 +1,14 @@
+local procedure UpgradeCustomerDiscountField()
+begin
+    if UpgradeTag.HasUpgradeTag(GetCustomerDiscountFieldTag()) then
+        exit;
+
+    Customer.SetLoadFields("Discount %");
+    if Customer.FindSet() then
+        repeat
+            Customer."Discount %" := 5;
```
JesperSchulz, line 9: This good fixture is not behaviorally equivalent to the bad one: it removes the `Discount % = 0` and nonblank posting-group guards and updates every customer. Microsoft's guidance also explicitly says to retain extra safety checks to avoid data corruption. Flatten by extracting the guarded update into a helper/early-exit shape while preserving both conditions; do not remove the conditions or turn each safety condition into an independently tagged migration.

JesperSchulz (changes requested):
The six open correctness threads from the previous pass still apply: mandatory API-page entity names, the nonexistent `Sales Line."Total Amount"` field, activation of price lists used by the good tests, SOAP-vs-OData wording for published codeunits, anti-pattern-first routing for rollback logging, and the current `PageType` enum/catalogue. I found four additional correctness gaps inline.

The exact-tree validators pass locally (0 errors, two existing keyword-count warnings; 318-article deterministic index; 34 fixtures across 17 domains), and all three reported checks are green. Those checks are structural and do not catch these AL/API/behavioral defects. The PR is also currently conflicting with `main` in `microsoft/skills/review/al-style-review.md`. Please address the ten focused items and rebase once; no broader editorial cleanup is requested.

MichaelDieringer:
Thanks Jesper. Addressed all ten focused items from the 2026-09-15 re-review.

**The six carried-over threads:**
- `api-page-least-privilege-write-access` fixtures: added the mandatory `EntityName`/`EntitySetName` properties (AL0485).
- `pages-must-not-contain-business-logic` fixtures: `Sales Line` has no `"Total Amount"` field — replaced with the real `"Line Amount"` (field 103).
- `test-feature-scenario-tags.good.al` and `test-one-when-per-test.good.al`: `CreatePriceHeader` leaves the list in Draft status, which price calculation ignores. Added `Validate(Status, Active)` + `Modify` before the sales line that depends on it — verified the `Status` field/enum against `PriceListHeader.Table.al`/`PriceStatus.Enum.al` in the BCApps clone.
- `exposed-objects-must-be-in-a-permission-set.md`: corrected — a published codeunit is a SOAP endpoint (deprecated), not OData; Page/Query are the OData object types. Now recommends API pages/queries for new integrations.
- `al-error-handling-review.md`: the cue selected on `Session.StartSession`, which only appears in the *compliant* fix — the bad fixture could never be worklisted. Re-cued on the actual risk shape (log insert around a failed `TryFunction`/`GetLastError*` path, then raise/propagate), with `StartSession` as a compliant discriminator instead.
- `page-design-must-match-bc-page-type-conventions.md`: fixed `Navigate` → `NavigatePage`, and noted the type list is a selected subset (not an exhaustive catalogue — `PromptDialog`/`ConfigurationDialog`/`UserControlHost`/`XmlPort` exist but are out of this article's scope).

**The four new gaps:**
- `release-must-update-app-version.md`: corrected — `id` is the app's stable identity, `version` identifies a release/code-state of it (was backwards).
- `defensive-vs-offensive-code-must-match-blast-radius.good.al`: added the missing `else` branch — a failed `Customer.Get()` was leaving the field at its prior/default value instead of the explicit chosen fallback the article claims to demonstrate.
- `bcpt-scenarios-must-be-app-specific.good.al`: `InitTest` and both measured sections were empty/comment-only. Filled in a real, self-contained header+line creation path so the fixture actually measures something.
- `upgrade-tag-logic-must-not-nest-deeply.good.al`: the flattened version had silently dropped both safety conditions from the bad fixture (`Discount % = 0`, nonblank posting group) instead of just removing nesting. Extracted the guarded update into a helper with both conditions preserved as early exits.

Also converted this PR's remaining plain-backtick sample references (16 articles) to the READ-convention markdown-link form, same fix already made on #156/#158.

Rebased onto current `main` (conflicts in `al-ui-review.md`, `al-style-review.md`, `al-upgrade-review.md` against merged upstream PRs — all additive, both sides' cues retained). All local validators pass: frontmatter (0 errors, the same two pre-existing keyword-count warnings), knowledge-index, knowledge-retrieval (351 articles / 589 samples), review-fixtures (108 cases / 20 leaf domains), skill-index.

JesperSchulz (changes requested):
Four merge-critical issues remain:

1. `pages-must-not-contain-business-logic.good.al` directly assigns `Sales Line."Line Amount"` and calls `Modify()`, bypassing validation and related amount/discount/VAT maintenance. The authoritative good sample can persist inconsistent Sales Lines; use a safe custom example or a supported validation/business routine.

2. `bcpt-scenarios-must-be-app-specific.good.al` depends on `Customer.FindFirst()` and a session-local counter for persistent keys. It fails in an empty company and collides on later BCPT sessions, so it is neither self-contained nor repeatable.

3. `upgrade-tag-logic-must-not-nest-deeply.md` treats business-data guards inside an upgrade loop as evidence for separate tagged upgrades. Microsoft's cited guidance requires such safety checks and demonstrates them inside the loop. Restrict this rule to nested tag gates or genuinely distinct migrations, excluding safety guards within one tagged migration.

4. `table-design-must-match-bc-table-type-conventions.md` routes essentially every new keyed table into nine non-exhaustive archetypes. Valid buffers, queues, logs, mappings, and working tables fall outside that list, creating systematic harmful redesign findings. Apply it only when context establishes one of the listed archetypes.

The prior API metadata, price activation, protocol wording, rollback routing, PageType, manifest, and fallback blockers otherwise appear resolved.

microsoft/knowledge/web-services/api-page-least-privilege-write-access.good.al
JesperSchulz: The API fixtures now include the required entity metadata. This discussion is resolved.

microsoft/knowledge/style/pages-must-not-contain-business-logic.good.al
JesperSchulz: The nonexistent field was replaced, so the original compilation concern is resolved. The latest review separately flags the safety of the replacement mutation and tracks that in the current change request.

microsoft/knowledge/testing/test-feature-scenario-tags.good.al
JesperSchulz: The price-list setup now activates and persists the header before price calculation. Resolved.

microsoft/knowledge/security/exposed-objects-must-be-in-a-permission-set.md
JesperSchulz: The guidance now distinguishes SOAP codeunit publication from OData/API page and query integrations. Resolved.

microsoft/knowledge/ui/page-design-must-match-bc-page-type-conventions.md
JesperSchulz: The PageType guidance now uses current names and is explicitly a selected conventional list rather than an exhaustive catalogue. Resolved.

microsoft/skills/review/al-error-handling-review.md
JesperSchulz: Routing now cues on transactional failure logging and treats `StartSession` as a compliant discriminator rather than the entry condition. Resolved.

microsoft/knowledge/error-handling/defensive-vs-offensive-code-must-match-blast-radius.good.al
JesperSchulz: The good fixture now assigns an explicit fallback when lookup fails. Resolved.

microsoft/knowledge/appsource/release-must-update-app-version.md
JesperSchulz: The manifest wording now preserves the stable app ID and correctly describes version as distinguishing releases/code states. Resolved.

microsoft/knowledge/testing/bcpt-scenarios-must-be-app-specific.good.al
JesperSchulz: The fixture now performs real app-specific work, resolving the original empty-scenario concern. The latest review separately tracks self-contained data and cross-run key repeatability.

JesperSchulz:
There has been significant progress here. I closed nine resolved or superseded discussions; the remaining review is now a bounded set of four concrete safety and precision fixes.

MichaelDieringer:
Thanks Jesper. All four fixed:

1. **\`pages-must-not-contain-business-logic\`**: replaced the real \`Sales Line\` example with a self-contained \`"Sample Order Line"\` table and switched the codeunit to \`Validate()\`/\`Modify(true)\` instead of a raw field assignment + \`Modify()\` — the "good" sample no longer bypasses a real document table's validation cascade regardless of which object it lives in.
2. **\`bcpt-scenarios-must-be-app-specific\`**: no longer assumes a customer exists (\`Customer.FindFirst()\` now falls back to creating one), and keys are generated from \`CreateGuid()\` instead of a session-local counter, so the scenario is self-contained and safe to run concurrently/repeatedly.
3. **\`upgrade-tag-logic-must-not-nest-deeply\`**: you're right, and this was a real conflation on our part — confirmed by re-reading Microsoft's own "Upgrading Extensions" page directly. "Keep tags simple by limiting nesting tags to two levels" is about nesting *tag checks*, and the page's own worked example nests a record loop with two business-data safety conditions inside one tagged procedure, matching the separate design guidance that explicitly *requires* such safety checks. Rewrote the Description/Best Practice/Anti Pattern to scope the rule to actual tag-check nesting or migrations blended under one tag, and rewrote both fixtures accordingly: \`good.al\` now nests safety conditions inside one migration's own loop (matching Microsoft's example) plus a second, genuinely separate migration as its own flat procedure; \`bad.al\` now shows the real anti-pattern — one tag's check nested inside another's guarded body.
4. **\`table-design-must-match-bc-table-type-conventions\`**: added an explicit scope note that the nine types aren't exhaustive, and narrowed the \`al-data-modeling-review.md\` cue to require positive evidence of one of the nine types (a naming suffix, key shape, or usage) rather than firing on any new table with a \`keys\` block.

Validators pass clean (frontmatter 0 errors, same 2 pre-existing keyword-count warnings; knowledge-index, knowledge-retrieval, review-fixtures all green).

JesperSchulz (changes requested):
One merge-critical inconsistency remains. The revised upgrade article now correctly permits business-data safety guards within one tagged migration, but `al-upgrade-review.md` still worklists nested “multi-branch business-data” or “separately-conditioned business” decisions. The reviewer can therefore still emit the exact false positive the article now excludes.

Please limit the cue to nested upgrade-tag checks or multiple functionally unrelated migrations under one tag, and explicitly exclude record loops and safety guards belonging to a single migration. The other three blockers from the latest review are resolved.

JesperSchulz (changes requested):
One merge-critical inconsistency remains. The revised upgrade article now correctly permits business-data safety guards within one tagged migration, but `al-upgrade-review.md` still worklists nested “multi-branch business-data” or “separately-conditioned business” decisions. The reviewer can therefore still emit the exact false positive the article now excludes.

Please limit the cue to nested upgrade-tag checks or multiple functionally unrelated migrations under one tag, and explicitly exclude record loops and safety guards belonging to a single migration. The other three blockers from the latest review are resolved.

JesperSchulz (changes requested):
One merge-critical inconsistency remains. The revised upgrade article now correctly permits business-data safety guards within one tagged migration, but `al-upgrade-review.md` still worklists nested “multi-branch business-data” or “separately-conditioned business” decisions. The reviewer can therefore still emit the exact false positive the article now excludes.

Please limit the cue to nested upgrade-tag checks or multiple functionally unrelated migrations under one tag, and explicitly exclude record loops and safety guards belonging to a single migration. The other three blockers from the latest review are resolved.

PR #158, opened 2026-09-04
3 AL/BC patterns from CURABIS's internal automated-testing training material

MichaelDieringer:
## Summary

Third batch from CURABIS ApS, mined from CURABIS's own internal Academy course book on automated/BDD testing in Business Central. Same format as #156/#157 (6-key frontmatter, Description/Best Practice/Anti Pattern, sibling `.good.al`/`.bad.al` samples, no fenced code in the .md).

- **data-modeling** (1): after posting a sales order with Ship and Invoice together, the resulting Item Ledger Entry carries the shipment's document number (`Sales Header."Last Shipping No."`), not the posted invoice number — a lookup filtered on the invoice number silently matches zero rows. Sales-specific; purchase combined posting (Receive+Invoice) uses different fields.
- **testing** (3): verify a field's actual rendered UI state with `TestPage` field `.Visible()`/`.Enabled()`, and its editability specifically with `.Editable()` opened via `OpenEdit()` — these are distinct states requiring different verification. Generate test fixture values via the right `LibraryUtility` helper for the guarantee actually needed: `GenerateRandomCodeWithLength`/`GenerateRandomCode20` when uniqueness against the real table must be verified, `GenerateGUID()` (untruncated, full-length fields only) or `GenerateRandomCode`/`GenerateRandomXMLText` for incidental values — not hardcoded literals or a truncated GUID.

Most of the source book turned out to be BDD philosophy and CURABIS process content that doesn't clear BCQuality's admission test — these 4 are the genuinely atomic, non-obvious AL-specific patterns that survived a full read-through, checked against the current `microsoft/knowledge/` corpus and CURABIS's own two prior PRs for overlap.

## Test plan


microsoft/knowledge/testing/use-generateguid-for-unique-test-fixture-values.md
```diff
@@ -0,0 +1,26 @@
...
+technologies: [al]
+countries: [w1]
+application-area: [all]
+---
+
+# Generate unique test fixture values with LibraryUtility.GenerateGUID()
+
+## Description
+
+A fixture helper that assigns a hardcoded literal to a primary-key or descriptive field collides the moment two tests, or two runs of the same test, create that fixture without cleanup, and a literal longer than the field allows raises a truncation or insert error. `LibraryUtility.GenerateGUID()` returns a value that is unique per call and long enough to guarantee no collision; paired with `CopyStr(..., 1, MaxStrLen(Field))` it fits any fixed-length `Code` or `Text` field safely.
```
JesperSchulz: `Library - Utility`.GenerateGUID() is not a GUID generator: current BCApps implements it as a `Code[10]` number-series value (`GU00000000` through `GU99999999`). Full returned values are sequentially unique within that series, but `CopyStr(..., 1, MaxStrLen(Field))` is unsafe for arbitrary shorter fields because it removes the changing rightmost digits; consecutive calls can then produce the same truncated value. Please constrain the guidance to fields that can hold all 10 characters (or prescribe a different strategy for shorter fields), and remove the absolute “guarantee no collision” / “any fixed-length field” claims.

microsoft/knowledge/testing/use-testpage-visible-enabled-to-verify-field-ui-state.md
```diff
@@ -0,0 +1,26 @@
...
+technologies: [al]
+countries: [w1]
+application-area: [all]
+---
+
+# Verify field visibility and editability with TestPage.Visible()/.Enabled()
+
+## Description
+
+A UI test codeunit does not need to inspect table or page properties indirectly to confirm a field is shown or editable under given conditions. The `TestPage` object exposes a `Visible()` and an `Enabled()` function on each field, reflecting the page's actual rendered state, callable directly from a `[Test]` procedure after `OpenView()`.
```
JesperSchulz: `Enabled()` and `Editable()` are distinct `TestField` methods. `Enabled()` gets the control's enabled state; `Editable()` gets its editable state. This sample opens the page with `OpenView()` and calls `Enabled()`, so it does not establish that the field is editable as the title, description, and assertion message claim. Either reframe this article strictly around visibility/enabled state, or use an appropriate page mode plus `Editable()` when the requirement is editability.

JesperSchulz (changes requested):
The Item Ledger Entry pattern checks out against current BCApps source: combined Ship+Invoice posting stamps the item journal line with the posted shipment number, while `LibrarySales.PostSalesDocument` returns `Last Posting No.` because the invoice branch overwrites its return-field selection.

I found two accuracy blockers in the testing guidance (called out inline) and one consumption gap: the current `al-testing-review` worklist has no `GenerateGUID`/`CopyStr` or `TestPage` `Visible`/`Enabled`/`Editable` cues or targeted checks, and `al-data-modeling-review` scopes relevance to setup/master-table surfaces without any Item Ledger Entry/posting cue. Please wire all three articles into their corresponding review-skill worklists so agents can actually select them; coordinate this with #155, whose contracts state that targeted checks cover every current article.

The CLA is also still pending. Because the PR says the material was mined from CURABIS's internal Academy course book, the contributor needs to use the applicable company/employer declaration, confirm authorization to contribute the derived material, and make any third-party-material disclosure required by the CLA.

MichaelDieringer:

microsoft/knowledge/testing/use-generateguid-for-unique-test-fixture-values.md
MichaelDieringer: Confirmed, GenerateGUID() is a Code[10] number-series value (GU00000000–GU99999999), not a real GUID. It returns the full 10 characters; the collision risk is on the caller side, truncating with CopyStr(..., 1, MaxStrLen(...)) cuts off the changing rightmost digits.

Beyond just narrowing scope: the same library already has the correct pattern for shorter or arbitrary-length fields, so we'll point to that instead of just restricting GenerateGUID()'s claim. GenerateRandomCode(FieldNo, TableNo) cuts from the left side instead (CopyStr(GenerateGUID(), 10 - FieldRef.Length + 1)), preserving the changing digits, and loops until RecRef.IsEmpty() to verify actual uniqueness against the real table — which GenerateGUID() alone never does. For arbitrary lengths not tied to a GUID at all.
GenerateRandomCodeWithLength/GenerateRandomXMLText(Length) generate true random per-character strings with the same table-uniqueness check. We'll rewrite the article to recommend these instead of prescribing manual truncation.

MichaelDieringer:

microsoft/knowledge/testing/use-testpage-visible-enabled-to-verify-field-ui-state.md
MichaelDieringer: Great catch too, rather than patch the sample, we're splitting this into two articles: the existing one stays scoped strictly to Visible()/Enabled() state (unchanged sample, editability claim dropped), and a new one covers Editable() specifically with OpenEdit() and its own scenario. 
Matches your suggestion, just doing both instead of picking one.

microsoft/knowledge/testing/use-generateguid-for-unique-test-fixture-values.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Best Practice
+
+For a field that holds the full 10 characters, assign `LibraryUtility.GenerateGUID()` directly. For a shorter or arbitrary-length field, use `LibraryUtility.GenerateRandomCode(FieldNo, TableNo)` (or `GenerateRandomCodeWithLength`/`GenerateRandomXMLText(Length)` for a specific length) instead of truncating a GUID yourself — these generate the value and verify it is actually unique against the target table, rather than relying on the number series alone.
```
JesperSchulz: These helpers do not share the claimed guarantee. `GenerateRandomCode` opens a **temporary** `RecordRef`, so its emptiness check does not inspect the target table. `GenerateRandomXMLText` performs no table lookup. `GenerateRandomCodeWithLength` does check the real table, but returns `Code[10]`, which constrains the meaningful length. Please document each helper's actual behavior and stop promising verified target-table uniqueness for all three.

microsoft/knowledge/testing/use-generateguid-for-unique-test-fixture-values.good.al
```diff
@@ -0,0 +1,19 @@
...
+{
+    var
+        LibraryUtility: Codeunit "Library - Utility";
+
+    procedure CreateCustomerType(var CustomerType: Record "Customer Type")
+    begin
+        CustomerType.Init();
+        // Code is shorter than GenerateGUID()'s 10 characters, so use
+        // GenerateRandomCode instead of truncating a GUID ourselves — it
+        // verifies uniqueness against the table rather than just returning
```
JesperSchulz: This repeats the false uniqueness guarantee from the article. `GenerateRandomCode` uses a temporary `RecordRef`; it does not verify the generated code against existing rows in `Customer Type`. Please rewrite the comment and choose a fixture strategy whose collision behavior matches the guarantee you actually want to teach.

microsoft/skills/review/al-testing-review.md
```diff
@@ -50,6 +50,8 @@ The following targeted checks cover every current `testing` article. Treat each
 - A permission-sensitive test uses `TestPermissions = Disabled`, claims to test a restricted user without `"Permissions Mock"`/`"Library - Lower Permissions"`, or declares `[TestPermissions(...)]` without applying that context — `permission-tests-must-lower-the-execution-context`.
 - Test fixture code manually calls `Init`/`Insert`, invents keys or prerequisite records, or bypasses available `LibrarySales`, `LibraryPurchase`, `LibraryERM`, `LibraryInventory`, `LibraryRandom`, or equivalent library codeunits — `use-library-codeunits-for-test-fixtures`.
 - `asserterror` is added or changed without a following `Assert.ExpectedError`, `Assert.ExpectedErrorCode`, or a purpose-built assertion such as `ExpectedTestFieldError` — `asserterror-needs-expectederror-and-code`.
+- Fixture code calls `LibraryUtility.GenerateGUID()` or `CopyStr` against it — `use-generateguid-for-unique-test-fixture-values`.
```
JesperSchulz: This cue primarily recognizes one compliant/helper-based shape and can miss the actual violations: hardcoded primary keys, truncated/generated values, and hand-built “unique” test values. It also lacks common TestPage/test-library relevance signals, allowing the reviewer to decide testing is not applicable before reaching these rules. Please cue on the anti-pattern and broaden relevance enough for the three new testing articles to be reachable.

microsoft/skills/review/al-data-modeling-review.md
```diff
@@ -46,6 +46,7 @@ A file enters the candidate worklist when its `keywords` intersect the extracted
 The following targeted checks cover every current `data-modeling` article. Treat each as a candidate-selection cue: when the signal appears in changed code, add the named article to the worklist and evaluate it in Action.
 
 - A `* Setup` table or its page changes singleton structure, uses a nonblank or generated key, permits insert/delete, uses a List page, or does not ensure the blank-keyed row exists — `setup-table-is-a-singleton`.
+- Code reads `Item Ledger Entry."Document No."` (or `"Last Shipping No."`/`"Last Posting No."`) after a combined Ship+Invoice sales/purchase post — `item-ledger-entry-document-no-follows-last-shipping-no`.
```
JesperSchulz: Please do not extend this sales rule to purchases. Purchase combined posting is Receive+Invoice and uses receiving fields such as `"Last Receiving No."`; the shipment/document-number behavior described by this article is specifically a sales path. Restrict this cue to sales unless a separate purchase rule is researched and documented.

JesperSchulz (commented):
Follow-up after `ab4ed50`: the Visible/Enabled scope and the Editable/OpenEdit distinction now look sound, and the Item Ledger Entry article is reasonable for combined **sales** Ship+Invoice posting. The three remaining substantive issues are inline.

Please also update the PR description: it still advertises the superseded `GenerateGUID() + CopyStr` recommendation and no longer describes the current change.

microsoft/knowledge/testing/use-generateguid-for-unique-test-fixture-values.md
MichaelDieringer: Fixed in 1d05811 — rewrote the Best Practice section to describe each `LibraryUtility` helper's actual behavior individually, verified against `LibraryUtility.Codeunit.al`: `GenerateRandomCode` opens the target table with `RecRef.Open(TableNo, true, CompanyName)` — `true` = temporary — so its own `IsEmpty()` check never touches real rows despite taking `TableNo`. `GenerateRandomXMLText` does no table lookup at all — it's a plain random-text generator. `GenerateRandomCodeWithLength`/`GenerateRandomCode20` open the real table (`false` = not temporary) and genuinely loop until `IsEmpty()`, but `GenerateRandomCodeWithLength` is declared `Code[10]` regardless of the requested length, so its meaningful length is capped at 10 characters. Rewrote the guidance to pick the helper by the guarantee actually needed instead of one blanket claim covering all of them.

microsoft/knowledge/testing/use-generateguid-for-unique-test-fixture-values.good.al
MichaelDieringer: Fixed in 1d05811 — switched the fixture to `GenerateRandomCodeWithLength(FieldNo, TableNo, MaxStrLen(Code))`, which does open the real (non-temporary) table, and rewrote the comment to attribute the verified-uniqueness guarantee to that call instead of to `GenerateRandomCode`.

microsoft/skills/review/al-testing-review.md
MichaelDieringer: Fixed in 1d05811 — rewired the cue to the actual anti-pattern shapes (a hardcoded literal, a hand-built "unique" value, or a GUID truncated for a field shorter than 10 characters) instead of only matching the compliant `GenerateGUID()+CopyStr` call, and broadened the token list (`Library - Utility`, `LibraryUtility`, `GenerateGUID`, `GenerateRandomCode`, `TestPage`, `.Visible(`, `.Enabled(`, `.Editable(`, `OpenNew`, `OpenView`, `OpenEdit`) so all three testing articles in this PR are actually reachable.

microsoft/skills/review/al-data-modeling-review.md
MichaelDieringer: Fixed in 1d05811 — restricted the cue to sales combined posting and added an explicit note that purchase combined posting is Receive+Invoice with different fields (`Last Receiving No.`), so it won't be worklisted from purchase-posting code.

MichaelDieringer:
@JesperSchulz Addressed in 1d05811 (pushed 2026-09-08): the 4 inline issues above, plus the PR description — updated to describe the actual current recommendation (pick the `LibraryUtility` helper by the guarantee needed) instead of the superseded blanket `GenerateGUID()+CopyStr` line. Ready for another look whenever you have time.

microsoft/knowledge/testing/use-generateguid-for-unique-test-fixture-values.md
```diff
@@ -0,0 +1,33 @@
...
+technologies: [al]
+countries: [w1]
+application-area: [all]
+---
+
+# Generate unique test fixture values with LibraryUtility helpers, not hardcoded literals
+
+## Description
+
+A fixture helper that assigns a hardcoded literal to a primary-key or descriptive field collides the moment two tests, or two runs of the same test, create that fixture without cleanup, and a literal longer than the field allows raises a truncation or insert error. `LibraryUtility.GenerateGUID()` is not a real GUID — it is a `Code[10]` number-series value (`GU00000000`–`GU99999999`) — and it returns the full 10 characters unshortened. Truncating it yourself with `CopyStr(..., 1, MaxStrLen(ShorterField))` for a field under 10 characters is unsafe: the changing digits sit at the right end and are exactly what gets cut off, so consecutive calls into a short field can produce the same truncated value. `GenerateGUID()` is only safe as-is for a field that holds the full 10 characters.
```
JesperSchulz, line 14: Please restrict the collision rule to primary/unique-key fields or values the test explicitly relies on as unique lookup identifiers. An ordinary descriptive field is not unique, so two rows with the same description do not collide on insert; deterministic descriptive text is also often required for exact assertions. The anti-pattern at line 31 and the testing review cue should be narrowed consistently so agents do not flag valid descriptive test data.

microsoft/knowledge/testing/use-generateguid-for-unique-test-fixture-values.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Best Practice
+
+For a field that holds the full 10 characters, assign `LibraryUtility.GenerateGUID()` directly. For a shorter field, do not truncate a GUID yourself — but also do not assume every `LibraryUtility` helper verifies uniqueness against the real table, because they don't all behave the same way:
+
+- `GenerateRandomCode(FieldNo, TableNo)` opens the target table as a **temporary** `RecordRef`, so its own emptiness check never inspects real rows — despite taking `TableNo`, it does not verify against the actual table. It's safe to use for its non-colliding-*within-a-single-test-run* value (derived from `GenerateGUID()`'s own number series), not for a guarantee against pre-existing or leftover data.
```
JesperSchulz, line 20: `GenerateRandomCode` cannot guarantee non-collision even within one test run for short fields. It takes the rightmost `FieldRef.Length` characters of the sequential `GenerateGUID()` value but does not remember earlier temporary results; a `Code[1]` value repeats every 10 calls, `Code[2]` every 100, and so on. Please remove the within-run guarantee and describe it consistently with line 25 as a finite short-field namespace with only a low collision chance. Keep the real-table helpers for cases that require verified uniqueness.

JesperSchulz (commented):
Thank you—the helper behavior, fixture implementation, review relevance, sales-only posting scope, and PR description are all much improved. This PR is very close.

Two factual points remain in the uniqueness article. We recognize that another round is inconvenient, but these claims directly determine which test code agents will flag and which helper guarantees they will trust, so they need to be precise. We have limited the feedback to those two points and omitted optional cleanup.

microsoft/knowledge/testing/use-generateguid-for-unique-test-fixture-values.md
JesperSchulz: Rechecked against current `LibraryUtility.Codeunit.al` at BCApps `42d69968a63e2fd951bba15e3dc51b50df932ffe`: this guarantee is still false. `GenerateRandomCode` keeps no history after opening a temporary `RecordRef`; its rightmost `n` digits repeat every `10^n` calls. Please describe it only as a finite short-field namespace with collision risk, and reserve verified uniqueness for the real-table helpers.

microsoft/knowledge/testing/use-generateguid-for-unique-test-fixture-values.md
JesperSchulz: Rechecked at `1d05811b3da1eaa4acefea533d2764bf62951d7b`: this is still unresolved. Please narrow the article description/anti-pattern and the testing worklist cue to primary/unique-key fields or values explicitly used as unique lookup identifiers. Deterministic descriptive values remain valid and should not be flagged.

JesperSchulz (changes requested):
Re-reviewed the exact latest head against current `main`, current BCApps test-library source, all prior threads, and the repository validators. The previously resolved helper/fixture/purchase-scope threads remain fixed; the two still-open uniqueness threads remain necessary.

Three additional blockers surfaced in the current tree:

1. `item-ledger-entry-document-no-follows-last-shipping-no.bad.al` calls `FindSet()` without consuming its optional Boolean. On an empty result AL raises a runtime error; it is not the silent zero-row behavior claimed by the article. Return/consume the Boolean in both fixtures (or correct the prose) so the sample demonstrates the claimed behavior.
2. `al-data-modeling-review.md` still declares the leaf `not-applicable` outside setup/master/key/numbering/block/audit surfaces (intro and outcome), which excludes the new sales-posting/Item Ledger Entry lookup despite its targeted cue. Add that surface to both scope clauses.
3. Current `main` requires READ-convention Markdown links to sibling samples. All eight new `See sample: `...`` references are plain code spans, so the exact virtual-merge `Test-ReviewFixtures.ps1` fails with “Sample is not linked by its article.” Convert each to ``[`name.good.al`](name.good.al)`` / bad equivalent.

Existing GitHub checks are green but predate the current-base validator. On the virtual merge, frontmatter and deterministic index checks pass; fixture validation fails at the links. After simulating only those eight link fixes, all 46 default cases pass, and a 50-case targeted run confirms all four new articles survive deterministic candidate ranking.

MichaelDieringer:
Thanks Jesper. Addressed all items from the 2026-09-15 re-review:

**The two still-open uniqueness threads:**
- Narrowed the collision rule (Description, Anti Pattern, and the `al-testing-review.md` worklist cue) to primary-key fields or values the test explicitly relies on as unique lookup identifiers. An ordinary descriptive field carries no such constraint, so a hardcoded or deterministic value there is no longer flagged.
- Corrected `GenerateRandomCode`: rechecked against current `LibraryUtility.Codeunit.al` — it opens its target table as a **temporary** `RecordRef`, which starts and stays empty, so its `repeat...until RecRef.IsEmpty()` always exits after one iteration and never retries, even within a single call. It returns the rightmost `FieldRef.Length` characters of `GenerateGUID()`'s sequential `GU00000000`–`GU99999999` series, so for a short field that window cycles (`Code[1]` repeats every 10 calls, `Code[2]` every 100). No longer describes it as collision-free at any scope — just a finite short-field namespace with low collision *chance*, distinct from the real table-checking helpers.

**The three new blockers:**
1. Both `item-ledger-entry-document-no-follows-last-shipping-no` fixtures called `FindSet()` without consuming its optional Boolean, which raises a runtime error on an empty result — the opposite of the article's own claimed "silently matches zero rows" behavior. Fixed with `if ItemLedgerEntry.FindSet() then;`, matching the existing idiom in `guard-database-reads.good.al`.
2. `al-data-modeling-review.md` still declared `not-applicable` outside setup/master/key/numbering/block/audit surfaces in both the intro and the outcome clause. Added dimension wiring, posting-routine structure, and Item Ledger Entry document-number lookups to both.
3. Converted all 8 plain-backtick sample references across the 4 new articles to the READ-convention markdown-link form.

Rebased onto current `main` (one conflict in `al-data-modeling-review.md`'s intro wording — merged, kept the upstream phrasing plus this PR's domain additions). All local validators pass: frontmatter, knowledge-index, knowledge-retrieval (337 articles / 565 samples), review-fixtures (108 cases / 20 leaf domains), skill-index.

JesperSchulz (changes requested):
One merge-critical contradiction remains in `al-testing-review.md`: it presents `GenerateRandomCode20` as compliant for a shorter field. That helper is intended for `Code[20]`; short fields can truncate the changing portion and make uniqueness retries pathological. This also contradicts the new article, which correctly scopes it to `Code[20]`.

Please route short fields to `GenerateRandomCodeWithLength` and reserve `GenerateRandomCode20` for `Code[20]` fields. The prior uniqueness, `FindSet`, scope, relevance, sample-link, description, and CLA blockers otherwise appear resolved.

microsoft/knowledge/testing/use-generateguid-for-unique-test-fixture-values.md
JesperSchulz: The article and cue are now restricted to key/explicit unique-lookup values and no longer flag ordinary deterministic descriptive data. Resolved.

microsoft/knowledge/testing/use-generateguid-for-unique-test-fixture-values.md
JesperSchulz: The false within-run guarantee is removed and the finite short-field collision risk is now described accurately. The latest review tracks one remaining helper-selection contradiction separately. Resolved.

JesperSchulz:
This is very close now. Both earlier discussion threads are resolved and closed; only the focused helper-selection contradiction from the latest review remains.

MichaelDieringer:
Thanks Jesper. Narrowed the \`al-testing-review.md\` cue — it no longer presents \`GenerateRandomCode20\` as an equal alternative to \`GenerateRandomCodeWithLength\` for a shorter field. Verified against \`LibraryUtility.Codeunit.al\`: \`GenerateRandomCode20\` truncates \`GenerateGUID()\`'s sequential value down to the field's length via \`PadStr\`, keeping the leftmost (slowest-changing) characters — so for a field shorter than 20, its retry loop can churn through the same truncated prefix for a long time, exactly the pathological-retries risk you flagged. \`GenerateRandomCodeWithLength\` is now the only compliant route for shorter fields; \`GenerateRandomCode20\` is scoped to \`Code[20]\`, matching what the article itself already said. Validators pass clean.

JesperSchulz (approved):
The remaining helper-selection contradiction is resolved: short fields now route to GenerateRandomCodeWithLength, while GenerateRandomCode20 is limited to Code[20]. No new merge-critical issue found.

PR #159, opened 2026-09-05
4 AL/BC testing patterns from an external BC testing expert's blog (Luc van Vugt, fluxxus.nl)

MichaelDieringer:
## Summary

Fourth batch from CURABIS ApS, this time mined from a source outside CURABIS itself: Luc van Vugt's fluxxus.nl, a long-running blog by a BC/NAV developer who authored a book on automated testing and ATDD in Business Central. The blog runs 2009–2023 and is mostly legacy classic-client/C-AL content; these 4 are the patterns that survived a full read-through, are still current in modern AL, and cleared BCQuality's admission test.

- `Confirm()` + `StrSubstNo`: a `[ConfirmHandler]`'s `Question` parameter receives the raw, unsubstituted template when `Confirm` is called with its own placeholder-substitution overload — the production code must build the string with `StrSubstNo` first for a handler-side assertion to see the real message.
- Table Relation Test's `OnAfterRemoveTableRelation` exclusion hook — verified directly against BCApps source (`codeunit 134926 "Table Relation Test"`, `src/Layers/W1/Tests/Misc/TableRelationTest.Codeunit.al`) rather than taken on the blog's word alone.
- Shared/lazy `Initialize()` fixture data needs an explicit `Commit()`, or it rolls back with the first test that creates it and silently vanishes for every test after.
- `Assert.IsFalse` for a boolean check, not `asserterror` wrapped around `Assert.IsTrue` — `asserterror` verifies that an error was raised, not the value under test.

Two other candidates found during mining were rejected: a `TestPermissions::InheritFromTestCodeunit` enum-value restriction (too thin/overlapping the existing permission-tests rule) and a suite-wide fixture-injection event pattern from 2018/2020 NAV content that no longer exists anywhere in current BCApps source (verified absent, not proposed).

## Test plan


JesperSchulz (changes requested):
Thanks for this — the mining quality is high. Verifying the Table Relation Test pattern against BCApps source instead of taking the blog's word for it is exactly the right instinct, and rejecting two candidates (one for no longer existing in current source) is the kind of restraint this corpus needs.

I re-verified all four claims independently. **All four are factually correct.**

| Claim | Verified against |
|---|---|
| `ConfirmHandler` receives the raw template | Reproduced in the source blog; corroborated by microsoft/ALAppExtensions#23935 |
| Table Relation Test exclusion hook | codeunit 134926, exact path; `OnAfterRemoveTableRelation` is an `[IntegrationEvent]` and `RemoveTableRelation` is **public** with the exact 5-arg signature |
| `Commit()` in a lazy `Initialize()` | 778 BCApps test files use this exact pattern |
| `Assert.IsFalse` over `asserterror` | `IsFalse` exists in both `Assert` and `Library Assert` (130002) |

The table-relation article is the strongest of the four: the pattern only works because `RemoveTableRelation` is public rather than local, and the bad sample's claim that passing `0` widens the delete is correct — `RemoveTableRelation` skips `SetRange` for any zero argument. Samples are independently written rather than copied, and each good/bad pair differs in exactly one thing.

Validation is green locally: frontmatter 0 errors, knowledge index deterministic at 304 articles, 34 review fixtures across 17 domains.

Requesting changes on the following.

---

### 1. CLA (blocking)

`license/cla` is still pending. Please sign it — but note this isn't purely procedural here. The CLA's *Originality of Work* clause asks that contributions derived from a third party be accompanied by the phrase "Submission containing materials of a third party:" followed by the third-party name and any known licenses. Since this batch is explicitly derived from a named author's blog, that clause is genuinely engaged rather than boilerplate. Please include it when you sign.

### 2. `commit-shared-test-fixture-inside-lazy-initialize.md` contradicts existing guidance

This is the one substantive issue. The existing `microsoft/knowledge/testing/transactionmodel-attribute-governs-test-transactions.md` states that applying `AutoRollback` to a test whose code calls `Commit` "produces a runtime error on the first Commit... the test does not complete." Your new article instructs the reader to call `Commit()` inside `Initialize()` under exactly that default.

Both files are `domain: testing`, `bc-version: [all]`, `countries: [w1]`, `application-area: [all]` — identical applicability with incompatible normative guidance, which is precisely the shape READ's conflict detection is meant to catch. It also has a live consequence: `al-testing-review.md` already tells agents to flag code that "calls `Commit` under `AutoRollback`", so as things stand an agent would flag the canonical BCApps fixture pattern.

The evidence says **your article is right and the existing one overstates**. `src/Layers/W1/Tests/ERM/ERMOnlineMappingSetup.Codeunit.al` commits inside `Initialize()`, declares no `TransactionModel` attribute at all, and passes — along with 777 other test files.

Since this PR is what surfaces the conflict, please correct the overstatement in `transactionmodel-attribute-governs-test-transactions.md` as part of this change, so the two articles agree.

### 3. `confirm-needs-strsubstno...` presents a platform bug as designed behavior

The article explains that substitution happens "only for the dialog a real user sees," which reads as intended platform behavior. It isn't — it's a reported defect (microsoft/ALAppExtensions#23935), raised by Luc van Vugt himself, with an internal bug filed at the time by @nikolakukrika.

That issue was closed as COMPLETED in February 2024, but the thread contains no fix confirmation and I found no public release-note evidence either way, so the current status is genuinely unclear. With `bc-version: [all]` and a Best Practice that tells agents to restructure **production** code, a silent platform fix would turn this into agents rewriting correct code for no reason.

Please reframe it as a known platform defect with the issue linked. @nikolakukrika — since you filed the internal bug and own this folder, could you confirm its disposition before we ship this as unbounded guidance?

Worth adding while you're in there: per the source blog, `Message`/`MessageHandler` substitutes correctly and only `Confirm` is affected. That asymmetry is both good evidence it's a bug and useful protection against an agent over-generalizing the rule to `MessageHandler`.

### 4. Minor — `table-relation-test-exclude-known-invalid-relations-via-event.md`

- Codeunit 134926 ships in BCApps' **test** app, so only consumers depending on the BC test libraries can subscribe. Worth stating explicitly.
- "walks every `TableRelation` field property in the app" is slightly off — it reads `Table Relations Metadata` filtered to `1..1999999999`, i.e. tenant-wide across installed apps, not just the current app.
- No token in `al-testing-review.md` covers `TableRelation` or `OnAfterRemoveTableRelation`, so this article may never surface in review. The other three are already reachable via `ConfirmHandler`, `asserterror`, `Commit`, and `Library Assert`. Happy to add the tokens separately if you'd rather keep this PR scoped to knowledge.

---

Note that `/microsoft/knowledge/testing/` is owned by @nikolakukrika, @ventselartur and @pchriste-microsoft-com, so this needs one of them to approve regardless of my review. Only `flag` and `guard` have run so far — the frontmatter and index workflows still need to go green in CI once a maintainer approves the run.

Nice work overall. Item 1 is quick; 2 and 3 are the ones that need real attention.

MichaelDieringer:

Submission containing materials of a third party: Luc van Vugt (fluxxus.nl) - blog content, no explicit license found.

MichaelDieringer:
Point 1: the fast one... done...

Point 2: Verified directly in ERMOnlineMappingSetup.Codeunit.al (codeunit 134915). No TransactionModel attribute at all, Commit() inside Initialize(), and manual cleanup via asserterror Error(...) at the end rather than relying on automatic rollback.
That's the actual mechanism: the "Commit causes an error" behavior is specific to the explicit AutoRollback attribute, not the undeclared default. 
We'll correct transactionmodel-attribute-governs-test-transactions.md's "Best Practice: Default to AutoRollback" wording so it stops conflating "no attribute" with "AutoRollback enforced," since that's what would make an agent flag this exact canonical pattern.

Point 3: Checked the issue directly. No fix confirmation visible in the thread either, just closed as completed with no comment. Will reframe as a known, unconfirmed platform defect with the issue linked, and add the Message/MessageHandler asymmetry as supporting evidence per your note.

Point 4: Fair, will note the test-app dependency explicitly, correct the "every TableRelation field property" wording to the actual tenant-wide Table Relations Metadata scope, and add the missing al-testing-review.md tokens in this PR rather than leaving it unreachable.

MichaelDieringer:

Submission containing materials of a third party: Luc van Vugt (fluxxus.nl) - blog content, no explicit license found.

JesperSchulz (commented):
Follow-up after `1e22b51`: CLA/disclosure now pass, and the Table Relation dependency and tenant-wide metadata wording are improved. The remaining transaction, applicability, fixture, and routing issues are inline.

Please also preserve durable provenance in the knowledge files themselves: each externally inspired article should identify its specific source and what was derived from it. A PR-level disclosure is useful for CLA review but will not accompany the rule when an agent consumes it later.

microsoft/knowledge/testing/commit-shared-test-fixture-inside-lazy-initialize.md
```diff
@@ -0,0 +1,26 @@
...
+technologies: [al]
+countries: [w1]
+application-area: [all]
+---
+
+# Commit shared fixture data created inside a lazy Initialize(), or later tests lose it
+
+## Description
+
+A test codeunit that creates master/setup data once, guarded by an `IsInitialized` flag, to avoid repeating expensive setup across many `[Test]` methods depends on that data surviving into every later test. Each `[Test]` method runs under `AutoRollback` by default, so data inserted during the first test's call to `Initialize()` rolls back at the end of that test. `IsInitialized` is a variable, not persisted data, so it still reads `true` on the next test — but the fixture rows it points to are already gone.
```
JesperSchulz: The core transaction explanation is incorrect. Microsoft's TransactionModel reference identifies `AutoCommit` as the default, and current BCApps tests call `Commit()` without an explicit attribute. Under AutoCommit, the method commits at its boundary without this added call; under AutoRollback, `Commit()` is rejected; under function-level `TestIsolation`, the fixture is rolled back between methods; and under codeunit-level isolation, the outer codeunit transaction already preserves it between methods. The explicit `Commit()` pattern in BCApps often establishes a rollback baseline inside a test, not this universal shared-fixture rule. Please remove or fundamentally rewrite this article around a demonstrated transaction/isolation configuration.

Reference: https://github.com/MicrosoftDocs/dynamics365smb-devitpro/blob/main/properties/properties/devenv-transactionmodel-property.md

microsoft/knowledge/testing/confirm-needs-strsubstno-before-confirmhandler-sees-substituted-text.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Anti Pattern
+
+Calling `Confirm('text %1', false, Value)` and then asserting the substituted text against `Question` inside a `[ConfirmHandler]`. `Question` holds the raw `'text %1'` template, so the assertion never matches the intended message.
+
+See sample: `confirm-needs-strsubstno-before-confirmhandler-sees-substituted-text.bad.al`.
+
+## Source
+
+Reported by Luc van Vugt: https://github.com/microsoft/ALAppExtensions/issues/23935 — an internal Microsoft bug was filed from that report; the issue's fix status is not confirmed as of this writing.
```
JesperSchulz: The linked issue is closed as **completed**, so this cannot remain an unconditional `[all]` workaround based on an “unconfirmed” fix. Please reproduce it on currently supported versions and set an accurate `bc-version`, or remove the rule if it no longer reproduces. The attribution is also inaccurate: Matjaž Šega opened microsoft/ALAppExtensions#23935; Luc van Vugt discussed/blogged it and credited Malcolm Gray. Please distinguish those roles instead of saying Luc reported the issue.

microsoft/knowledge/testing/confirm-needs-strsubstno-before-confirmhandler-sees-substituted-text.bad.al
```diff
@@ -0,0 +1,9 @@
+codeunit 50140 "Sample Confirm Usage"
+{
+    procedure ConfirmDeletion(RecordCount: Integer): Boolean
+    var
+        ConfirmMsg: Label 'Do you want to delete %1 records?';
+    begin
+        exit(Confirm(ConfirmMsg, false, RecordCount));
```
JesperSchulz: This fixture only shows a `Confirm` call; it has no `[ConfirmHandler]` or assertion against `Question`, so it cannot demonstrate the alleged failure condition. If the rule survives a current-version reproduction, the bad fixture must include the handler assertion that fails and the good fixture must show the same assertion passing after explicit substitution.

microsoft/knowledge/testing/table-relation-test-exclude-known-invalid-relations-via-event.md
```diff
@@ -0,0 +1,26 @@
...
+technologies: [al]
+countries: [w1]
+application-area: [all]
+---
+
+# Exclude a known-valid TableRelation exception via OnAfterRemoveTableRelation
+
+## Description
+
+Codeunit 134926 "Table Relation Test" (shipped in BCApps' test app — only consumers that depend on the BC test libraries can subscribe to it) reads Table Relations Metadata tenant-wide across every installed app, not just the current one, and fails the moment a related field's type or length doesn't match what the relation requires — the related field must match the largest related field's length, and its type must match (except a field may relate to both `Code` and `Text`, which resolves to `Text`). A field with a legitimate, intentional relation shape has no per-field override in its own object definition; the check runs with no built-in escape hatch.
```
JesperSchulz: The relevant compatibility validation in `Table Relation Test` is `[Scope('OnPrem')]`. Please state that boundary in the applicability/prose and fixtures; otherwise agents may recommend subscribing to a test surface that is not available in a cloud-targeted test app.

microsoft/knowledge/testing/use-assert-isfalse-not-asserterror-for-boolean-checks.md
```diff
@@ -0,0 +1,26 @@
...
+technologies: [al]
+countries: [w1]
+application-area: [all]
+---
+
+# Use Assert.IsFalse to check a boolean result, not asserterror around Assert.IsTrue
+
+## Description
+
+`asserterror` exists to assert that a statement raises a runtime error; it is not a general-purpose way to invert a boolean check. Wrapping `asserterror Assert.IsTrue(SomeFunc(), Msg)` to verify that `SomeFunc()` returns `false` tests whether `Assert.IsTrue`'s own error-raising behavior fired, not the value `SomeFunc()` actually returned.
```
JesperSchulz, line 14: The local recommendation is reasonable, but this shape also matches the existing generic `asserterror-needs-expectederror-and-code` rule. Without explicit precedence/suppression, an agent can load both and suggest adding error assertions instead of replacing the construct with `Assert.IsFalse`. Please wire the specialized rule to win for `asserterror Assert.IsTrue(BooleanExpression, ...)` and document the non-overlapping scope.

microsoft/skills/review/al-testing-review.md
```diff
@@ -50,6 +50,9 @@ The following targeted checks cover every current `testing` article. Treat each
 - A permission-sensitive test uses `TestPermissions = Disabled`, claims to test a restricted user without `"Permissions Mock"`/`"Library - Lower Permissions"`, or declares `[TestPermissions(...)]` without applying that context — `permission-tests-must-lower-the-execution-context`.
 - Test fixture code manually calls `Init`/`Insert`, invents keys or prerequisite records, or bypasses available `LibrarySales`, `LibraryPurchase`, `LibraryERM`, `LibraryInventory`, `LibraryRandom`, or equivalent library codeunits — `use-library-codeunits-for-test-fixtures`.
 - `asserterror` is added or changed without a following `Assert.ExpectedError`, `Assert.ExpectedErrorCode`, or a purpose-built assertion such as `ExpectedTestFieldError` — `asserterror-needs-expectederror-and-code`.
+- A `[ConfirmHandler]` asserts `Question` against a substituted message from a `Confirm` call using its placeholder-substitution overload — `confirm-needs-strsubstno-before-confirmhandler-sees-substituted-text`.
+- A shared/lazy `Initialize()`-style fixture helper calls `Commit()` — `commit-shared-test-fixture-inside-lazy-initialize`.
```
JesperSchulz: This selects the article when `Commit()` is present—the shape the article calls compliant—and can completely miss the actual missing-Commit anti-pattern. Candidate cues must recognize potential violations, not just examples of the proposed fix. If the transaction article is retained after correction, cue on lazy initialization plus shared state/fixture creation and evaluate its TransactionModel/TestIsolation context explicitly.

JesperSchulz:
Correction to my own review. The transaction analysis in my first round was wrong, and I would rather retract it explicitly than leave it standing next to a follow-up that contradicts it.

### What I got wrong

In the first round I wrote that "the evidence says **your article is right and the existing one overstates**", and asked you to correct `transactionmodel-attribute-governs-test-transactions.md` as part of this PR. That was backwards. You should not have been asked to make that edit.

### The fact

`AutoCommit` is the documented default transaction model for a test method — [TransactionModel Property](https://github.com/MicrosoftDocs/dynamics365smb-devitpro/blob/main/properties/properties/devenv-transactionmodel-property.md): "**AutoCommit** is the default value."

Worth noting for anyone following this thread: the [TransactionModel *attribute*](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/attributes/devenv-transactionmodel-attribute) page — the one you naturally reach for — never names a default at all. That is probably how this got past both of us.

### Why my evidence never supported my conclusion

`codeunit 134915 "ERM Online Mapping Setup"` declares no `TransactionModel` attribute, so it runs under `AutoCommit`. It is not an `AutoRollback` test. The existing article's claim is explicitly scoped to *applying* `AutoRollback`, so an attribute-less codeunit cannot show that it overstates.

I counted 778 files sharing the pattern and treated that as confirming the article's stated *mechanism*. Prevalence of a practice is not evidence for the reason given for it — and that distinction is most of what this corpus is for.

The giveaway was in the file I quoted. It ends with a deliberate `asserterror Error(RollBackMessage)`, which only makes sense if the `Commit()` is establishing a rollback baseline — the reading in my follow-up — rather than protecting a shared fixture from a per-test rollback that never happens.

### What this means for the PR

1. Please treat my first-round request to edit `transactionmodel-attribute-governs-test-transactions.md` as **withdrawn**. Your edit is not wrong as far as it goes — absence of the attribute genuinely is not equivalent to `AutoRollback` — but it was made on a bad instruction from me, it dropped the article's default recommendation, and it still never says what an attribute-less method actually does.
2. As it stands this PR ships two `domain: testing` articles that disagree. The new one says test methods run "under `AutoRollback` by default"; the edited one says absence is not `AutoRollback`; neither names `AutoCommit`. Whatever survives the rewrite, please make both agree and state the default explicitly.

### One more from my first round

I wrote that the ConfirmHandler issue was "raised by Luc van Vugt himself". microsoft/ALAppExtensions#23935 was opened by Matjaž Šega; Luc blogged it and credited Malcolm Gray. My follow-up corrected that inline, but the error originated in the round that presented a table asserting I had independently verified all four claims and that all four were factually correct.

That table was overconfident. One of the four has since been ruled incorrect, and one of its supporting attributions was wrong. I would rather say so plainly, since "verified against source" is the bar this repository asks contributors to meet.

### Unaffected

Nothing here touches the other three articles, and the table-relation facts I checked in the first round all hold: codeunit 134926, `RemoveTableRelation` public with the five-argument signature, zero arguments widening the delete, and the tenant-wide `1..1999999999` filter.

Sorry for the churn. This one was my error, not yours.

microsoft/knowledge/testing/commit-shared-test-fixture-inside-lazy-initialize.md
MichaelDieringer: Fixed in 0b08c2f — the article was rewritten around the same correction you made independently in your retraction: `AutoCommit` is the documented default `TransactionModel`, not `AutoRollback`. The rewritten article now explains the actual mechanism — `Commit()` inside `Initialize()` protects the fixture from the test method's own later deliberate `asserterror`-based rollback (the BCApps `ERMOnlineMappingSetup` pattern), not from an `AutoRollback` default that doesn't exist for an attribute-less method — and makes the fixture's survival across test methods explicitly dependent on `TestIsolation` (`Disabled`/`Codeunit` survive across methods, `Function` does not).

microsoft/knowledge/testing/confirm-needs-strsubstno-before-confirmhandler-sees-substituted-text.md
MichaelDieringer: Deleted the article and both fixtures in 0b08c2f — microsoft/ALAppExtensions#23935 is closed as completed (Feb 2024) and I couldn't reproduce or pin the behavior to any currently supported version, so an unconditional `[all]` workaround isn't defensible. (The attribution issue is moot since the article's gone, but noted for the record: you're right that Matjaž Šega opened the issue and Luc credited Malcolm Gray.)

microsoft/knowledge/testing/confirm-needs-strsubstno-before-confirmhandler-sees-substituted-text.bad.al
MichaelDieringer: Same resolution as the .md file — deleted in 0b08c2f rather than fixed, since the underlying platform bug is reported closed and I couldn't reproduce it to build a fixture that demonstrates a real current failure.

microsoft/knowledge/testing/table-relation-test-exclude-known-invalid-relations-via-event.md
MichaelDieringer: Fixed in 0b08c2f — added the `[Scope('OnPrem')]` boundary to both the applicability prose and the Best Practice section, verified directly against codeunit 134926's `ValidateFieldRelationCompatibility()` test method. Made explicit that subscribing to this exclusion hook has no effect for a cloud-targeted test app, since the test itself never runs there.

microsoft/knowledge/testing/use-assert-isfalse-not-asserterror-for-boolean-checks.md
MichaelDieringer: Fixed in 0b08c2f — added a Scope section wiring `use-assert-isfalse-not-asserterror-for-boolean-checks` to win specifically for `asserterror Assert.IsTrue(BooleanExpression, ...)`, and documented the non-overlapping boundary with the generic `asserterror-needs-expectederror-and-code` rule.

microsoft/skills/review/al-testing-review.md
MichaelDieringer: Fixed in 0b08c2f — rewired the shared-fixture cue to the actual missing-Commit anti-pattern (lazy initialization + shared fixture creation, evaluated against the method's TransactionModel/TestIsolation context) instead of firing on `Commit()`'s presence, which was the compliant shape. Also added the missing cue for `use-assert-isfalse-not-asserterror-for-boolean-checks` with precedence over the generic asserterror rule, and removed the cue for the deleted `confirm-needs-strsubstno` article.

MichaelDieringer:
Appreciate the correction — no worries on the churn. 0b08c2f reflects it: the `transactionmodel-attribute-governs-test-transactions.md` edit from round 1 is superseded, both testing articles now state the `AutoCommit` default explicitly and agree with each other, and neither treats an attribute-less method as equivalent to `AutoRollback`. Also resolved the Matjaž Šega/Malcolm Gray attribution issue by deleting the article that had it (see the thread above) — the underlying platform bug is reported closed and I couldn't reproduce it on a current version, so there's no longer an unconditional workaround making that claim.

MichaelDieringer:
@JesperSchulz Addressed this round in 0b08c2f (pushed 2026-09-08): the 6 inline issues above, the transaction-model contradiction from your retraction (both articles now state and agree on the `AutoCommit` default), and the in-file Source provenance you asked for — each surviving externally-inspired article now names its specific fluxxus.nl post and states what was independently verified vs. taken from it, not just disclosed at the PR level. Ready for another look whenever you have time.

microsoft/knowledge/testing/commit-shared-test-fixture-inside-lazy-initialize.md
JesperSchulz: Thanks—the rewrite correctly establishes `AutoCommit` as the default and narrows the explicit `Commit()` to the deliberate rollback-sentinel pattern. Three connected issues remain:

1. `TestIsolation = Disabled` is still listed alongside `Codeunit` as a supported recommendation. Under `Disabled`, the early commit is never rolled back at codeunit completion and permanently contaminates the database. The recommended configuration should be `Codeunit`; mention `Disabled` only with an explicit verified final teardown.
2. The good fixture's `asserterror Error(RollBackMsg)` is still selected by the generic `asserterror-needs-expectederror-and-code` rule, so the new canonical pattern violates another active rule. Please add explicit precedence/exclusion for this deterministic rollback sentinel.
3. The fixtures create no persisted fixture data, contain only one test, and show no runner/isolation context, so they cannot demonstrate either the cross-method failure or the proposed fix. Please make the sample executable with a later test and explicit Codeunit isolation context.

With those changes, the transaction model and the agent guidance should align.

microsoft/knowledge/testing/table-relation-test-exclude-known-invalid-relations-via-event.md
```diff
@@ -0,0 +1,30 @@
...
+technologies: [al]
+countries: [w1]
+application-area: [all]
+---
+
+# Exclude a known-valid TableRelation exception via OnAfterRemoveTableRelation
+
+## Description
+
+Codeunit 134926 "Table Relation Test" (shipped in BCApps' test app — only consumers that depend on the BC test libraries can subscribe to it) reads Table Relations Metadata tenant-wide across every installed app, not just the current one, and fails the moment a related field's type or length doesn't match what the relation requires — the related field must match the largest related field's length, and its type must match (except a field may relate to both `Code` and `Text`, which resolves to `Text`). A field with a legitimate, intentional relation shape has no per-field override in its own object definition; the check runs with no built-in escape hatch. The validation test method itself is `[Scope('OnPrem')]`: it only runs from an on-premises test surface, not from a cloud-targeted test app, so this whole exception mechanism — and the check it works around — is only reachable where that test can actually execute.
```
JesperSchulz, line 14: The length rule needs one qualification. Current `Table Relation Test` requires exact length when the metadata contains an unconditional relation. For conditional-only relations, it permits the source field to be longer than the largest related field and fails only when the source is shorter. Please document those two cases separately; otherwise agents will recommend exclusions or schema changes for conditional relations that the standard test accepts.

JesperSchulz (commented):
Thank you for the substantial rewrite. The Confirm rule removal, TransactionModel correction, OnPrem scope, Boolean-rule precedence, routing, and per-article attribution address most of the previous feedback.

The remaining transaction concerns are tracked in the existing conversation, so I am not opening duplicate threads for them. One additional factual qualification is needed below because it changes which table relations the article tells agents to reject. We recognize the extra round is inconvenient; this is limited to behavior that materially affects the rule's correctness. The branch also currently conflicts with `main` in the TransactionModel article and testing review skill and will need to be updated before merge.

microsoft/knowledge/testing/table-relation-test-exclude-known-invalid-relations-via-event.md
JesperSchulz: Rechecked `0b08c2f` against the current `ValidateFieldRelation` implementation in codeunit 134926. The description still states exact length/type globally. Please distinguish the actual branches: an unconditional relation requires exact maximum related-field length and exact resolved type; conditional-only relations fail length only when the source is shorter, and when the required type is `Code`, both `Code` and `Text` source fields are accepted. Capturing both qualifications now avoids agents rejecting shapes the standard test accepts. Leaving this unresolved.

microsoft/knowledge/testing/commit-shared-test-fixture-inside-lazy-initialize.md
JesperSchulz: Rechecked `0b08c2f` against the current TransactionModel/TestIsolation docs and the rewritten fixtures. The three follow-ups are still present: `Disabled` remains a recommended shared-fixture configuration even though it never rolls changes back; the deliberate `asserterror Error(...)` rollback sentinel is still selected by the generic ExpectedError rule; and the fixtures still create no persisted row, have no later test, and show no runner/isolation context, so they cannot demonstrate the failure or fix. Please recommend `Codeunit` isolation (`Disabled` only with explicit verified final teardown), add sentinel precedence/exclusion, and make the fixture executable and cross-method. Leaving this unresolved.

JesperSchulz (changes requested):
Re-reviewed the full latest diff and every prior thread. The deleted Confirm rule, provenance/CLA attribution, OnPrem boundary, narrow exclusion API, Boolean article scope, and candidate cues are otherwise sound. The exact-head frontmatter, index determinism (303 articles), and 34 fixtures/17 domains all pass locally; `flag`, `guard`, and CLA pass.

Changes still needed:

1. **Shared fixture:** resolve the three items left in the existing thread—recommend `TestIsolation = Codeunit` (not bare `Disabled`), exclude the deliberate rollback sentinel from the generic ExpectedError rule, and replace the non-demonstrating fixtures with persisted data + a later test + explicit runner/isolation context.
2. **Table Relation Test:** document both conditional-only exceptions from codeunit 134926: source length may exceed the maximum related length, and a required `Code` relation accepts a `Code` or `Text` source. Unconditional relations remain exact-length/exact-type.
3. **Deterministic routing:** `al-testing-review.md` currently excludes every `asserterror Assert.IsTrue/IsFalse` Boolean-returning shape from the generic rule, while the new article's Scope says the generic rule still owns cases where the guarded call itself is expected to error. Qualify the cue so the specialized rule wins only when the construct solely inverts the returned Boolean.
4. Resolve the current conflict with `main` in `transactionmodel-attribute-governs-test-transactions.md`, and update the stale PR title/summary from four patterns to the three surviving ones (remove the deleted Confirm claim).

These are the remaining correctness/mergeability issues; no additional round is intended beyond verifying those exact fixes.

MichaelDieringer:
Thanks Jesper. Addressed all four items from the 2026-09-15 re-review.

**1. Shared fixture (`commit-shared-test-fixture-inside-lazy-initialize`):**
- Recommend `TestIsolation = Codeunit`, not `Disabled` as an equal option — `Disabled`'s own documentation says "tests are not isolated from each other" (nothing ever rolls back), so a fixture this pattern commits is permanent database contamination unless something else tears it down. `Disabled` is now mentioned only alongside that explicit-teardown requirement.
- Added precedence in `al-testing-review.md` so the deliberate end-of-test `asserterror Error(...)` rollback sentinel isn't also caught by the generic `asserterror-needs-expectederror-and-code` rule.
- Rewrote both fixtures to actually demonstrate the pattern: `CreateSharedFixtureData` now inserts a real `Item` record instead of an empty comment, a second `[Test]` method depends on that fixture surviving into it, and both fixtures include an explicit `Subtype = TestRunner` / `TestIsolation = Codeunit` runner codeunit.

**2. Table Relation Test length/type rule:** verified `ValidateFieldRelation` in codeunit 134926 directly and split the article into the two branches the source actually has — a field with any *unconditional* relation needs exact length and exact resolved type; a field whose relations are all *conditional* only fails on being shorter (longer is accepted) than the largest related field, and when the required type is specifically `Code`, a `Text` source passes too — a tolerance that's conditional-only and doesn't extend to a required `Text`.

**3. Deterministic routing:** already resolved in the prior round per your read — no further change needed there.

**4. Rebase + stale title:** rebased onto current `main` (one conflict in `transactionmodel-attribute-governs-test-transactions.md` — upstream had already linked its sample references, ours added a `## Source` section; merged both). Updated the PR title and summary to "3 AL/BC testing patterns" and removed the deleted Confirm/StrSubstNo claim from the description.

Also converted the 3 remaining plain-backtick sample references in this PR to the READ-convention markdown-link form, same fix as #156/#157/#158.

All local validators pass clean: frontmatter (0 errors, 0 warnings), knowledge-index, knowledge-retrieval (336 articles / 563 samples), review-fixtures (108 cases / 20 leaf domains), skill-index.

JesperSchulz (changes requested):
Four merge-critical issues remain:

1. In `al-testing-review.md`, the generic ExpectedError route excludes every `asserterror Assert.IsTrue/IsFalse` around a Boolean-returning call, while the specialized route accepts only constructs used solely to invert the Boolean. A test expecting the guarded Boolean function itself to raise therefore falls through both routes. Narrow the exclusion to the same inversion-only condition.

2. The rollback-sentinel exception exists only in skill prose while `asserterror-needs-expectederror-and-code.md` remains unconditional. False-positive exclusions are normative knowledge and must be encoded there so every consumer avoids the harmful recommendation.

3. `commit-shared-test-fixture-inside-lazy-initialize.good.al` hand-rolls an Item with `Init`, an invented key, and `Insert(true)`, which another active rule correctly flags. Use `LibraryInventory.CreateItem` so the canonical fix is clean under the full ruleset.

4. `table-relation-test-exclude-known-invalid-relations-via-event.good.al` and its bad pair reference undefined `Sample Header` and `Sample Setup` tables. The samples are not self-contained or compilable; define minimal tables or use existing symbols.

JesperSchulz:
The transaction-model rewrite is materially stronger. I left two discussions open because they still contain active contract details; the latest review makes the remaining fixes explicit and finite.

MichaelDieringer:
Thanks Jesper. All four fixed:

1. Narrowed \`al-testing-review.md\`'s generic ExpectedError exclusion for \`asserterror Assert.IsTrue/IsFalse\` to the same inversion-only condition the specialized rule already requires — a test expecting the guarded Boolean-returning call itself to raise now stays owned by \`asserterror-needs-expectederror-and-code\`.
2. Encoded the rollback-sentinel exception directly in \`asserterror-needs-expectederror-and-code.md\`'s Anti Pattern section, not just the skill's routing prose.
3. \`commit-shared-test-fixture-inside-lazy-initialize\`'s fixtures now use \`LibraryInventory.CreateItem\` instead of hand-rolled \`Item.Init()\`/\`Insert(true)\`.
4. \`table-relation-test-exclude-known-invalid-relations-via-event\`'s fixtures now declare minimal \`"Sample Setup"\`/\`"Sample Header"\` tables inline instead of referencing undefined symbols.

Validators pass clean.

JesperSchulz (changes requested):
One merge-critical fixture issue remains. The new `Sample Header."Category Code"` field has no `TableRelation`, so the good subscriber's targeted `RemoveTableRelation` removes no metadata. The bad sample also has no second valid relation whose coverage its table-wide removal would destroy. The pair therefore demonstrates neither the recommended exclusion nor its anti-pattern.

Please add an actual relation that creates the targeted metadata row and another valid relation on the same table that remains checked by the good subscriber but is incorrectly removed by the bad subscriber. The four blockers from the latest review are otherwise resolved.

MichaelDieringer:
Thanks Jesper. Fixed — you're right, "Category Code" had no \`TableRelation\` at all, so the good subscriber's \`RemoveTableRelation\` call targeted metadata that never existed. Added a real \`TableRelation = "Sample Setup"."Primary Key"\` on "Category Code" (the one known exception to exclude) and a second, ordinary self-referencing relation, \`"Parent No." -> "Sample Header"."No."\`, with no exception. The good fixture's targeted \`RemoveTableRelation(..., 10, ...)\` now removes only the first; the bad fixture's table-wide \`RemoveTableRelation(..., 0, 0, 0)\` now demonstrably also strips the second — so the pair actually shows both the recommended exclusion and its anti-pattern. Validators pass clean.

microsoft/knowledge/testing/commit-shared-test-fixture-inside-lazy-initialize.md
JesperSchulz: The current tree resolves all three follow-ups: it recommends `Codeunit` isolation, limits `Disabled` to verified teardown, gives the rollback sentinel explicit precedence over the generic ExpectedError rule, and includes persisted fixture data, a later test, and an explicit Codeunit-isolated runner. Resolved.

microsoft/knowledge/testing/table-relation-test-exclude-known-invalid-relations-via-event.md
JesperSchulz: The article now distinguishes unconditional from conditional-only relation checks, including the different length rules and the conditional Code/Text tolerance. Resolved.

JesperSchulz (approved):
The final fixture blocker is resolved: the good sample removes one real relation metadata row while preserving a second valid relation, and the bad sample demonstrably removes both. The earlier transaction, routing, knowledge-contract, fixture-library, and table-relation semantics issues are also resolved. No new merge-critical issue found.

PR #160, opened 2026-09-07
Clarify page field caption and tooltip inheritance

JesperSchulz:
## Summary

Correct guidance that caused agents to add redundant `Caption` and `ToolTip` properties to bound page fields.

- Explicitly recognize table-field caption inheritance and discourage duplicate page-level captions.
- Document table-field tooltip inheritance from BC24/runtime 13.0, retaining page-level guidance for older runtimes, unbound controls, and context-specific overrides.
- Require inspection of source fields before reporting missing properties; unavailable source definitions are not proof of a defect.
- Align the related UI article and good/bad AL examples with the corrected guidance.

## Validation

- Knowledge frontmatter and structure validator: no errors or warnings.
- Existing review fixture validation: 34 cases across 17 leaf domains passed.
- `git diff --check`: passed.

## References

- [Microsoft Learn: ToolTip property and inheritance](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/properties/devenv-tooltip-property)
- [Microsoft Learn: Caption property](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/properties/devenv-caption-property)

### DIFF
--- a/microsoft/knowledge/style/caption-required-on-page-fields.bad.al
+++ b/microsoft/knowledge/style/caption-required-on-page-fields.bad.al
@@ -9,16 +9,22 @@ page 50253 "Sample Caption Bad"
         {
             group(General)
             {
-                field("Customer No."; Rec."No.")
+                Caption = 'General';
+                field(CustomerNoValue; CustomerNoValue)
                 {
                     ApplicationArea = All;
+                    ToolTip = 'Specifies the customer number to look up.';
                 }
                 field("Customer Name"; Rec.Name)
                 {
                     ApplicationArea = All;
                     Caption = '';
+                    ToolTip = 'Specifies the customer name.';
                 }
             }
         }
     }
+
+    var
+        CustomerNoValue: Code[20];
 }
--- a/microsoft/knowledge/style/caption-required-on-page-fields.bad.al
+++ b/microsoft/knowledge/style/caption-required-on-page-fields.bad.al
@@ -19,7 +19,7 @@ page 50253 "Sample Caption Bad"
                 {
                     ApplicationArea = All;
                     Caption = '';
-                    ToolTip = 'Specifies the customer name.';
+                    ToolTip = 'Specifies the customer name shown on sales documents.';
                 }
             }
         }
--- a/microsoft/knowledge/style/caption-required-on-page-fields.good.al
+++ b/microsoft/knowledge/style/caption-required-on-page-fields.good.al
@@ -1,7 +1,36 @@
+// BC24 / runtime 13.0 or later for table-field tooltips.
+table 50252 "Sample Caption Source"
+{
+    Caption = 'Caption Source';
+    DataClassification = CustomerContent;
+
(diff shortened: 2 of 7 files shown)


microsoft/knowledge/style/caption-required-on-page-fields.bad.al
```diff
@@ -9,16 +9,22 @@ page 50253 "Sample Caption Bad"
...
+                field(CustomerNoValue; CustomerNoValue)
                 {
                     ApplicationArea = All;
+                    ToolTip = 'Specifies the customer number to look up.';
                 }
                 field("Customer Name"; Rec.Name)
                 {
                     ApplicationArea = All;
                     Caption = '';
+                    ToolTip = 'Specifies the customer name.';
```
nikolakukrika: I think we should put more information, than just repeating the field name. We can autogenerate these if needed, IMO they are not bringing much value. Could be one of the rules to nudge agent to generate a better descriptive names.

nikolakukrika: Do we need a knowledge article on this? I assume it should recover on the error and do the right thing?

JesperSchulz: That's actually a good point. Even though this is the bad code sample... 😉 But we should add that in the knowledge article.

JesperSchulz: That's a point worth discussing. Would we want the agent to understand how to write decent AL code from the get-go, or should it first try, then fail and learn based on compiler rules, and then succeed?

nikolakukrika (approved):

microsoft/knowledge/style/caption-required-on-page-fields.bad.al
JesperSchulz: For a purely mechanical missing-property diagnostic, compiler/analyzer feedback can guide recovery. The reason to keep this article is different: the original rule caused agents to add redundant page-level captions even when inheritance already supplied them. Those additions compile successfully, so there is no compiler error to recover from, and they prevent later table-caption changes from reaching the page.

Added that distinction to the article in 11aefa9. Its purpose is to prevent an observed false positive, not duplicate the compiler rulebook. Useful tooltip wording is another aspect that merely satisfying a presence check cannot establish.

JesperSchulz: Agreed. Updated the tooltip knowledge article to require information beyond the caption: what the value is used for, expected values/units, or the effect of changing it, grounded in the actual implementation rather than invented behavior. Shared help should still live on the table field, not be repeated across pages.

Also improved the tooltip on this line and the companion examples. The caption bad sample now demonstrates only the intended caption problem; a weak but non-empty tooltip is explicitly distinguished from a missing tooltip. Changes are in 11aefa9.

PR #161, opened 2026-09-08
AL methods limited during write transactions (RunModal, Codeunit.Run)

MichaelDieringer:
## Summary
One new `performance` article, `al-methods-limited-during-write-transactions.md`, documenting the platform restriction behind the runtime error *"The following AL methods are limited during write transactions because one or more tables will be locked: Form.RunModal, Codeunit.Run, Report.RunModal, XmlPort.RunModal."* — one explicit line per method with its exact condition (Page.RunModal never; Report/XmlPort.RunModal only with the request page suppressed; Codeunit.Run only when the return value is unused), why the guard exists, and how to structure code so it never triggers.

**Why the Microsoft layer:** this is platform-enforced behavior, not a team convention. Verified against: Microsoft Learn (`Codeunit.Run` transaction semantics — "you must commit first"); a live reproduction on Business Central 26 (2026-09-07, `Item.Insert()` then `Page.RunModal(Page::"Customer Card")` — unrelated table — fails on the RunModal line; message quoted verbatim); Microsoft's own Base Application, which follows the `Commit(); Page.RunModal(...)` pattern in `ActivityLog.Table.al`, `DocumentSendingProfile.Table.al`, `PaymentServiceSetup.Table.al` and others; and microsoft/AL#5452 for the pre-2019 wording. Only the Codeunit.Run leg is documented on Learn; the RunModal legs exist only as the runtime error text — which is exactly why an agent gets this wrong without the file.

**Relationship to existing articles:** `codeunit-run-requires-prior-commit-inside-transaction.md` keeps ownership of the Codeunit.Run leg (cross-referenced, not duplicated). `avoid-user-prompts-inside-transactions.md` keeps the prompts the platform *allows* inside a write transaction (Confirm/StrMenu — which therefore silently hold locks); the two words "modal page" are removed from its list, because that case is refused with a runtime error, not stalled.

**Wiring:** `al-performance-review.md` gains `Page.RunModal`/`Report.RunModal`/`XmlPort.RunModal`/`UseRequestPage` tokens and one deterministic worklist cue with exclusions (RunModal before every write; request page suppressed), routing Codeunit.Run in that position to its existing owner — per #155's coverage contract.

**Samples** reference real objects only: page 428 "Shipping Agents", table 291 "Shipping Agent" (`Code[10]`), Sales Header fields 21 "Shipment Date" and 105 "Shipping Agent Code".

## Test plan
- [x] `validate_frontmatter.py --root .` — 0 errors, 0 warnings
- [x] `Build-KnowledgeIndex.ps1` — 301 articles, deterministic
- [x] `Test-ReviewFixtures.ps1` — 34 cases / 17 leaf domains
- [x] CLA signed on this PR (`company="CURABIS ApS"`)
- [ ] Domain owner review (`microsoft/knowledge/performance/`)


MichaelDieringer:

JesperSchulz (changes requested):
Requesting changes for one AL API/routing correctness issue.

The current AL API is `Xmlport.Run`, not `XmlPort.RunModal`. Microsoft Learn exposes `Xmlport.Run(Integer [, Boolean] [, Boolean] [, var Record])`; there is no XMLport `RunModal` method. Likewise, `UseRequestPage(false)` is a report-instance method; XMLports use the `UseRequestPage = false;` object property or the static `Xmlport.Run` `RequestWindow` argument. Please update the article's normative bullet, Best Practice/Anti Pattern wording, keywords, and `al-performance-review` tokens/cue to route actual `Xmlport.Run` calls. Keep `XmlPort.RunModal` only in the quoted/explained legacy runtime message. As written, the skill misses real AL XMLport calls and teaches an API that does not compile.

The Page fixture pair and remaining transaction claims check out against current Learn/BCApps. All three exact-head validators pass (frontmatter; deterministic 301-article index; 34 fixtures/17 leaves). Existing checks are green. The head currently conflicts with `main` in `al-performance-review.md` after #148; when rebasing, retain both the current job-queue additions and this corrected transaction cue.

MichaelDieringer:
Thanks for catching this — you're right, `XmlPort.RunModal` doesn't exist. Fixed (rebased onto `main` and force-pushed):

- Replaced `XmlPort.RunModal` with the real API, `Xmlport.Run(Integer [, RequestWindow: Boolean] [, Boolean] [, var Record])`, throughout the Description, the `al-performance-review.md` worklist cue, and the token list. Verified against Microsoft Learn's `Xmlport.Run` and `XMLport data type` pages — the XMLport data type has no `RunModal` method, static or instance.
- Corrected `UseRequestPage(false)`: that's a `Report` instance method only. For XMLports the equivalent is the `RequestWindow` argument to `Xmlport.Run`, or the `UseRequestPage = false;` object property (XMLports have no instance `UseRequestPage` method).
- Added a note explaining that the platform's own error message also says `XmlPort.RunModal`/`RequestForm` — same kind of legacy phrasing already called out for `Form.RunModal`, not a second real API.
- Added Learn citations for `Xmlport.Run` and `UseRequestPage` to the Source section.

While rebasing I also noticed the two `See sample:` links used plain backtick text instead of the READ-convention markdown-link form, which was failing your `Knowledge-Retrieval.ps1` check independently of this issue — fixed that too.

Rebased onto current `main` to pick up #148's job-queue additions to `al-performance-review.md`; both sets of worklist tokens/cues are retained. All four local validators pass (frontmatter, knowledge-index, knowledge-retrieval, review-fixtures).

JesperSchulz (changes requested):
One merge-critical routing gap remains. The article correctly covers request-page-enabled `Report.Run` and `Report.RunModal` after writes, but `al-performance-review.md` tokens/cues recognize only `Report.RunModal`. A failing `Modify(); Report.Run(..., true, ...)` path is therefore not reliably worklisted.

Add `Report.Run` to deterministic routing and exclude calls where the request page is suppressed (`Report.Run(..., false, ...)` or `UseRequestPage(false)`). The prior XMLport API/property corrections otherwise appear resolved.

JesperSchulz:
Nearly there: the prior XMLport/API corrections are sound. The latest review has only one remaining deterministic-routing gap for request-page-enabled Report.Run.

MichaelDieringer:
Thanks Jesper. Added \`Report.Run\` alongside \`Report.RunModal\` to \`al-performance-review.md\`'s token list and targeted cue (same request-page-suppressed exclusion), and named both methods explicitly in the knowledge article's Description bullet — they share the identical RequestWindow-blocking-dialog mechanism, differing only in whether the report instance is cleared afterward, so the guard and the routing now treat them the same. Frontmatter and knowledge-index validators pass clean.

JesperSchulz (approved):
The remaining routing gap is resolved. Request-page-enabled Report.Run is now routed deterministically, with suppressed request-page forms excluded. No new merge-critical issue found.

PR #163, opened 2026-09-08
knowledge: improve review precision from BCApps PR 7546-neg-dc4dbb701f46583f3b48adea8c1962c2cffd5865 feedback

gggdttt:
## Summary

Improves BCQuality knowledge based on maintainer thumbs-down feedback from explicit BCApps PR review runs.

## Source feedback

- https://github.com/microsoft/BCApps/pull/7546#discussion_r3796791415

## Validation

- Frontmatter and article structure validation completed.

Generated by the BC-ALAgentsInternal self-improvement workflow.

Pending offline evaluation; human review is required.

### DIFF
--- a/microsoft/knowledge/performance/avoid-get-inside-loop-on-large-table.md
+++ b/microsoft/knowledge/performance/avoid-get-inside-loop-on-large-table.md
@@ -13,9 +13,11 @@ application-area: [all]
 
 A `Get` or `FindFirst` against another persistent table inside a loop can produce an N+1 access pattern: one outer query followed by repeated inner lookups. Server and primary-key caches can satisfy some `Get` calls, so a source-level `Get` is not proof of one SQL round-trip. The concern is an unbounded loop whose lookup keys are not known to repeat or remain cached.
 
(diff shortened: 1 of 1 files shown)


JesperSchulz (changes requested):
Requesting changes because the source feedback has been over-generalized.

- The new text says inner cardinality caps *total lookup cost* and outer cardinality is irrelevant. Every qualifying outer row still executes `Get`; the [BC cache](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/administration/optimize-sql-data-access#business-central-server-data-caching) only avoids SQL on a cache hit. Likewise, a nonblank foreign-key guard may skip no rows. These are severity factors, not unconditional low-severity exclusions.
- `Allocation Account`, `Payment Terms`, and `No. Series` have no enforced few-hundred-row limit in their AL table definitions. The BCApps thread supports downgrading that specific finding and not demanding an invasive join, but not a universal "thousands of rows" threshold. Please key the rule to bounded loop/cardinality, expected distinct-key reuse/cache evidence, and measured hotness, then add/update a paired fixture or evaluation that exercises the new small-inner/guard exclusion. The current pair still covers only the large `Item` case, and the performance evaluation is pinned to another article.

The exact merge tree passes frontmatter, deterministic index/retrieval, review-contract, and all 46 review-fixture validations; all six PR checks pass and GitHub reports the branch mergeable.

gggdttt:
Closing this generated candidate as part of the reviewed cleanup for ADO 648887.

The originating automatic cycle did not produce completed offline-evaluation evidence for this candidate. Closing the pending candidate does not constitute acceptance of its knowledge change.

The source branch, commits and historical workflow evidence are retained. Any future proposal needs review and fresh evidence tied to its exact revisions; this closure does not authorize another evaluation run.

PR #175, opened 2026-09-10
5 AL/BC patterns: document distribution (Report Selections, Document Sending Profile, Find Entries, TransferFields)

MichaelDieringer:
## Summary

Five rules about Business Central's document distribution architecture, verified against BCApps source and Microsoft Learn:

- **custom-document-dispatch-must-not-bypass-report-selections** — a codeunit that hardcodes a report ID and builds its own email instead of registering through `Report Selections` loses per-account layout overrides and "Document Layouts" discoverability.
- **document-print-and-email-actions-call-report-selections-directly** — a document's own Print/Email actions call `Report Selections` directly; `Document Sending Profile` is scoped specifically to the combined Post-and-Send action, not general print/email.
- **extend-find-entries-navigate-for-new-document-types** — registering a custom table with Find Entries (`page 344 Navigate`) requires both `OnAfterFindRecords` and `OnBeforeShowRecords`; either alone leaves a result row with nothing to open it.
- **extend-report-selection-usage-for-new-document-types** — a new `Report Selection Usage` value needs both `ReportSelections.InsertRecord` and a subscriber to the Customer/Vendor "Document Layouts" filter events, or the per-account override UI never lists the new document type.
- **transferfields-mirrored-fields-must-match-type-and-length** — a field mirrored across a `TransferFields` posting cascade (e.g. `Sales Header` → `Sales Invoice Header`) must match type and length exactly; a length-only mismatch compiles and posts cleanly until a value finally exceeds the shorter definition.

Each article cites specific BCApps source (file, procedure, and in most cases line number) and, where applicable, Microsoft Learn. Wired into `al-data-modeling-review.md`'s worklist cues so all five are reachable in review. Added a disambiguation note on the TransferFields article distinguishing it from the existing `transferfields-skip-type-mismatch-can-drop-data.md` — that article covers `SkipFieldsNotMatchingType` masking a *type* mismatch; this one covers the *length* dimension, which that parameter does not affect at all.

## Test plan


MichaelDieringer:

JesperSchulz (changes requested):
Requesting changes on `83f041b6624b965c061e05bc36cbdc1926399be2`.

The exact-tree validators pass (frontmatter; deterministic 305-article index; 34 fixtures/17 leaves), but these are necessary before the rules are safe to execute:

1. **Owning-skill reachability is contradictory.** `al-data-modeling-review` still scopes relevance/not-applicable/outcome to setup/master/key/numbering/block/audit surfaces, and its initial tokens omit all five new signal families. The appended worklist cues therefore do not make document actions, Navigate subscribers, report-selection registration, or posting-cascade extensions deterministically reachable. Extend the entry gate/scope and prove each route.
2. **The Document Sending Profile rule and bad fixture do not match current BCApps semantics.** Ordinary `Sales Invoice Header.PrintRecords`/`EmailRecords` use stateless `DocumentSendingProfile.TrySendToPrinter`/`TrySendToEMail`, not direct `Report Selections`; the article names only `Purchase Header` as an exception. Conversely, the bad sample calls `Send` on a blank record. `Send` does not load the customer's assigned profile, so this sample does not make the outcome depend on that profile—it simply has all sending options at their default `No`. Permit the stateless helpers and load a configured/default customer profile in the actual anti-pattern.
3. **The Report Selection Usage rule is overbroad and incomplete.** Requiring both customer and vendor filter subscribers even for a one-sided document is contradicted by current `ReportSelectionHandlerCZZ`, which partitions sales usages to the customer page and purchase usages to the vendor page. Also, filter subscribers alone only affect **Copy from Report Selection**; full Document Layouts display/edit support requires extending the page-facing usage enum and handling its map/validate events, as the cited CZC implementation itself does. Scope to the applicable counterparty and either include those pieces or narrow every claim to the copy action.
4. **Evaluation does not exercise any new rule.** The generated data-modeling control still selects `check-blocked-in-referencing-code-not-in-master`; all five new good/bad pairs are only existence/ranking inputs. Add deterministic positive/clean coverage that demonstrates the corrected owning-skill routes and semantics.

JesperSchulz (changes requested):
The current head is unchanged from the prior review, and four merge-critical blockers remain:

1. The five rules are only targeted cues in `al-data-modeling-review.md`; its entry gate/not-applicable definition can reject document actions, Navigate subscribers, report-selection registration, and posting cascades before those cues run.

2. `document-print-and-email-actions-call-report-selections-directly.md` incorrectly rejects supported stateless `TrySendToPrinter`/`TrySendToEMail` usage. Its bad sample calls `Send` on a blank profile, so the no-op is not caused by the assigned customer profile as claimed.

3. `extend-report-selection-usage-for-new-document-types.md` requires both customer and vendor subscribers for one-sided documents and claims Document Layout support without the page-facing usage-enum mapping/validation integration. Its good sample contains only the two copy-filter subscribers.

4. None of the five rules has deterministic positive/clean evaluation coverage; the default data-modeling pair still selects an unrelated alphabetical article.

The conflict with current `main` is separate, but resolution must preserve current `folder-path` support.

JesperSchulz:
The remaining work is now a concrete four-item checklist plus the main-branch conflict. Once those are addressed, the next pass can stay tightly focused on merge readiness.

MichaelDieringer:
Thanks Jesper — apologies for how long this one sat. All four merge-critical items fixed:

1. **Entry-gate/scope**: `al-data-modeling-review.md`'s `not-applicable` definition, object/token lists, and worklist now explicitly cover document actions, `Navigate` subscribers, Report Selection registration, price-calculation/price-source extensibility, `TransferFields` posting-cascade mirroring, and barcode/font-provider usage. Traced each of the 9 slugs through the (extended) gate to confirm none are rejected upstream of their cue.

2. **document-print-and-email-actions-call-report-selections-directly**: you were right on both counts. Verified `SalesInvoiceHeader.Table.al`'s `PrintRecords`/`EmailRecords` call `TrySendToPrinter`/`TrySendToEMail` on a local record that's never `Get`'d — a legitimate stateless path, now explicitly permitted. And confirmed the bad fixture's blank-record `Send` call doesn't depend on any assigned profile; reworked it to load a real one via `GetDefaultForCustomer` first, matching the real anti-pattern.

3. **extend-report-selection-usage-for-new-document-types**: confirmed `ReportSelectionHandlerCZZ` partitions strictly by counterparty (sales → customer page, purchase → vendor page) versus `ReportSelectionHandlerCZC`, which genuinely needs both for Compensation. Rule now scopes to the applicable single counterparty by default, and added the page-facing usage-enum (`Custom Report Selection Sales`/`Report Selection Usage Vendor`) map/validate events alongside the filter subscription — the filter alone only wires "Copy from Report Selection", not full Document Layouts support, as you noted.

4. **Evaluation coverage**: noticed `review-fixtures.json` has since grown a real `articles[]` override convention (used by finance/scm/query/reporting/style) since this PR was opened — used that instead of inventing a parallel mechanism. All 9 new pairs are now exercised.

Also caught independently: a stale field-citation in `custom-document-dispatch-must-not-bypass-report-selections` (`Custom Report Layout Code` is field 7, not part of the 19–26 email-config range), and converted all 9 articles' sample links to the markdown-link form `Knowledge-Retrieval.ps1` requires.

Rebased onto current `main` (conflicts in `al-data-modeling-review.md` and `evaluation/README.md`, both resolved keeping both sides — `folder-path` support and the two new cues preserved).

Also folded in 4 new patterns while I was in here (price-calculation-handler activation, price-source/document-subset enum sync, price-source recalculation trigger, barcode font-provider usage) — same domain, same review skill, seemed better as one coherent PR than a second one right behind it. Title/description updated accordingly.

All local validators pass: frontmatter 0/0, review-fixtures 126/20 domains, knowledge-index 342/575, skill-index 19 leaves.

JesperSchulz (changes requested):
The original four blockers are addressed, but five merge-critical defects remain in the added price/barcode material:

1. `activate-new-price-calculation-handler-via-onfindsupportedsetup.md` and its review cue require every handler setup to be `Default := true`. Detailed Price Calculation Setup can select a registered non-default handler by code; default is required only for fallback selection.
2. `new-price-source-must-add-candidate-and-trigger-recalculation.good.al` calls `UpdateUnitPriceByField` without first calling `PlanPriceCalcByField`, so the method exits and the canonical fix recalculates nothing. Use the full planned-field sequence or a dedicated handler recalculation routine.
3. The barcode article universally requires `ValidateInput`, but `Barcode Font Provider 2D` does not expose that API. Split 1D (`ValidateInput` + `EncodeFont`) from 2D (`EncodeFont`).
4. The same rule rejects all manual delimiters, but `*value*` can be valid Code 39 encoding when the input/font/checksum requirements permit it. Flag demonstrably invalid or mismatched encoding rather than construction categorically.
5. Two expected-clean evaluation fixtures are not self-contained: the price-handler fixture references undefined `Sample Price Calc - Special`, and the Find Entries fixture references undefined sample table/page objects. Add minimal definitions or use existing symbols.

MichaelDieringer:
Thanks Jesper — all five fixed:

1. **`Default := true` absoluteness**: confirmed against `PriceCalculationMgt.Codeunit.al`'s `FindSetup` — it's genuinely two-stage: it first tries `PriceCalculationDtldSetup.FindSetup` (no `Default` condition at all, matches on Source Group/No./Asset Type/No.), and only falls back to the `Enabled`/`Default`/`Method`/`Type`/`Asset Type` filter if that misses. Softened the article and its worklist cue to reflect that `Default` only governs the fallback branch.

2. **`UpdateUnitPriceByField` no-op**: confirmed — `FieldCausedPriceCalculation` starts at 0 and `UpdateUnitPriceByField` exits immediately unless a prior `PlanPriceCalcByField` call set it to match. Switched the fixture to the public `UpdateUnitPrice` wrapper (`ClearFieldCausedPriceCalculation` + `PlanPriceCalcByField` + `UpdateUnitPriceByField` in one call), which is also what real BCApps code uses for this exact "recalculate from an unrelated field's OnValidate" shape (`ItemReferenceManagement.Codeunit.al`).

3. **Barcode 1D/2D split**: confirmed `Barcode Font Provider 2D` has no `ValidateInput` member — only `EncodeFont` (and `GetSupportedBarcodeSymbologies`). Split the article's Best Practice accordingly.

4. **Code 39 delimiter**: you're right that `*` is a legitimate Code 39 start/stop character, not inherently wrong. Reframed the anti-pattern around demonstrable mismatch — verified `IDA1DCode39Encoder.Codeunit.al`'s actual encoded output wraps in `(`/`)`, not literal `*`, so a hand-built `'*'+value+'*'` is mismatched with the paired production font specifically, which is the real defect, not manual construction per se.

5. **Undefined eval-fixture objects**: found and fixed both — a `"Sample Price Calc - Special"` codeunit referenced but never declared (added a real `implements "Price Calculation"` stub), and a `"Sample Posted Document Header"` table/page referenced but never declared in the Find Entries fixtures (declared both, self-contained, matching the convention already used elsewhere in this PR).

All validators pass: frontmatter 0/0, review-fixtures 126/20 domains, knowledge-index 342/575, skill-index 19 leaves.

PR #177, opened 2026-09-10
Add retention policy knowledge to the privacy domain

JeremyVyska:
Business Central's retention policy feature had no coverage in BCQuality — the only mention anywhere in the corpus was a passing clause in the data-classification article. Extension-owned tables that accumulate rows (activity logs, integration logs, archives) routinely ship with no policy at all, and the two mistakes behind that are both LLM-invisible: registration that only runs in install code, and registration mistaken for a policy.

Two articles in `microsoft/knowledge/privacy/`, each with `.good.al` / `.bad.al` companions:

**`register-owned-log-tables-for-retention-policies.md`** — the engine only deletes from tables in the allowed-tables list, and an extension may register only tables it owns. Register from install *and* upgrade code, because install code does not run when an existing installation moves to a new version; guard with `IsAllowedTable` plus an upgrade tag; use `MandatoryMinRetenDays` where audit needs a floor, and locked filters via `AddTableFilterToJsonArray`. Anti-pattern: a hand-rolled `DeleteAll` against a hard-coded window, invisible to the Retention Policy Log and unadjustable by the administrator.

**`ship-a-default-retention-policy-setup.md`** — `AddAllowedTable` only makes a table selectable. Nothing is deleted until a `Retention Policy Setup` record exists, names a period, and is enabled, so ship one pre-created and disabled, gated by an upgrade tag — the shape codeunit 3907 "Retention Policy Installer" uses for `Retention Policy Log Entry`. Anti-pattern in both directions: registering and stopping, or inserting the setup enabled so the app deletes tenant data on a schedule nobody approved.

Platform claims are cited in `## References` against the BC admin documentation and the Retention Policy module in microsoft/BCApps. Layer choice follows the current `skills/write.md` rule: `privacy` is owned by `al-privacy-review`, so its corpus lives in the Microsoft layer regardless of authorship.

Checks: `validate_frontmatter.py` reports 0 errors / 0 warnings; `Test-ReviewFixtures.ps1` passes at 34 cases across 17 leaf domains (the convention-selected privacy fixture is unchanged).

Deliberately out of scope, noted for a later PR: `Apply Retention Policy` deletes at most 10,000 records per run shared across all policies, so a table gaining more than that per day never converges. That is a throughput concern and belongs in `performance`, not here.

https://claude.ai/code/session_011gvTjm746MtJEVWeRTbG46

### DIFF
--- a/microsoft/knowledge/privacy/register-owned-log-tables-for-retention-policies.md
+++ b/microsoft/knowledge/privacy/register-owned-log-tables-for-retention-policies.md
@@ -1,0 +1,33 @@
+---
+bc-version: [all]
+domain: privacy
+keywords: [retention-policy, allowed-tables, addallowedtable, reten-pol-allowed-tables, log-table-growth, mandatory-minimum-retention, install-upgrade-codeunit]
+technologies: [al]
+countries: [w1]
+application-area: [all]
+---
+
+# Extension-owned log tables must be registered as retention-policy allowed tables
+
+## Description
+
+The retention policy engine only ever deletes from tables that appear in its allowed-tables list, and an extension may register only tables it owns — it cannot add a base application table or a table from another extension. Registration is a call to `Codeunit "Reten. Pol. Allowed Tables".AddAllowedTable`, passing the table ID and the field number of the Date or DateTime field that ages each record (`SystemCreatedAt` is the usual choice). Until that call has run in a company, the table cannot be selected on the **Retention Policies** page at all, so an activity log, integration log, or archive table the extension writes to grows with no supported way for an administrator to trim it. The registration is stored per company and is not part of the table's metadata — it exists only because install or upgrade code put it there.
+
+## Best Practice
+
+Register every table the extension owns that accumulates rows over time: activity and audit logs, integration and API request logs, archived documents. Call a shared routine from both the install codeunit (`OnInstallAppPerCompany`) and an upgrade codeunit (`OnUpgradePerCompany`), because install code does not run when an existing installation moves to a new version — registration added only to install code never reaches tenants that already have the app. Guard the routine with `IsAllowedTable` and an upgrade tag so repeated runs are idempotent. Pass `MandatoryMinRetenDays` when the data must survive a minimum period for audit or support reasons; the platform then rejects any shorter period an administrator configures. When only a subset of rows should ever expire, build the filter with `AddTableFilterToJsonArray` and pass it to the `AddAllowedTable` overload that takes a `JsonArray` — a filter added as locked cannot be removed later by the administrator.
+
+Registering a table only makes it eligible; the policy itself is a separate concern, covered by [`ship-a-default-retention-policy-setup.md`](ship-a-default-retention-policy-setup.md).
+
+See sample: [`register-owned-log-tables-for-retention-policies.good.al`](register-owned-log-tables-for-retention-policies.good.al).
+
+## Anti Pattern
+
+An extension-owned log table with no `AddAllowedTable` call anywhere in the app, cleaned instead by hand-rolled code — a job queue codeunit or scheduled task running `DeleteAll` against a hard-coded date window. The deletion happens outside the **Retention Policy Log**, the administrator has no page on which to lengthen, shorten, or disable it, and the table is invisible during a data-retention review. The same defect in slower form is registration performed only in the install codeunit: new tenants are covered, every existing tenant stays unregistered after the upgrade.
+
+See sample: [`register-owned-log-tables-for-retention-policies.bad.al`](register-owned-log-tables-for-retention-policies.bad.al).
+
+## References
+
+- [Clean up data with retention policies](https://learn.microsoft.com/en-us/dynamics365/business-central/admin-data-retention-policies), section *Include your extension in a retention policy*.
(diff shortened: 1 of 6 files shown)


microsoft/knowledge/privacy/register-owned-log-tables-for-retention-policies.md
```diff
@@ -0,0 +1,33 @@
+---
+bc-version: [all]
```
JesperSchulz, line 2: Retention policies and these System Application APIs were introduced with Business Central 2020 release wave 2 (v17), so `[all]` makes both articles applicable to versions where the objects do not exist. Please scope both articles to `bc-version: [17..]`, consistent with how this corpus versions other platform features.

microsoft/knowledge/privacy/register-owned-log-tables-for-retention-policies.md
```diff
@@ -0,0 +1,33 @@
...
+
+## Best Practice
+
+Register every table the extension owns that accumulates rows over time: activity and audit logs, integration and API request logs, archived documents. Call a shared routine from both the install codeunit (`OnInstallAppPerCompany`) and an upgrade codeunit (`OnUpgradePerCompany`), because install code does not run when an existing installation moves to a new version — registration added only to install code never reaches tenants that already have the app. Guard the routine with `IsAllowedTable` and an upgrade tag so repeated runs are idempotent. Pass `MandatoryMinRetenDays` when the data must survive a minimum period for audit or support reasons; the platform then rejects any shorter period an administrator configures. When only a subset of rows should ever expire, build the filter with `AddTableFilterToJsonArray` and pass it to the `AddAllowedTable` overload that takes a `JsonArray` — a filter added as locked cannot be removed later by the administrator.
```
JesperSchulz, line 18: Install and upgrade coverage is necessary, but not sufficient. The Retention Policies pages raise `Reten. Pol. Allowed Tables.OnRefreshAllowedTables`; current System Application, Base Application, Shopify, Email, and Performance Profiler installers all subscribe and re-run registration with a force path. This sample's upgrade-tag guard makes it exit after the first run, so it cannot participate in that refresh path. Please add the event subscriber and separate “refresh registration” from “one-time setup/tag” behavior, as the platform installers do.

microsoft/knowledge/privacy/ship-a-default-retention-policy-setup.good.al
```diff
@@ -0,0 +1,48 @@
...
+        end;
+
+        UpgradeTag.SetUpgradeTag(DefaultPolicyTag());
+    end;
+
+    local procedure SixMonthRetentionPeriod(): Code[20]
+    var
+        RetentionPeriod: Record "Retention Period";
+    begin
+        RetentionPeriod.SetRange("Retention Period", RetentionPeriod."Retention Period"::"6 Months");
```
JesperSchulz, line 33: Please use the public `Codeunit "Retention Policy Setup".FindOrCreateRetentionPeriod(...)` helper here instead of reimplementing it. This implementation can fail if a `Retention Period` record already uses code `SIX MONTHS` for a different enum value: the filtered `FindFirst` finds nothing, then `Insert` collides on the existing code. The public helper is specifically intended to find or safely create the requested period and is what current Shopify, Performance Profiler, and Financial Report code uses.

JesperSchulz (changes requested):
These are valuable additions: the distinction between allowed-table registration and an actual policy is important, the opt-in default is responsible, and the main platform claims align with the current retention-policy module. The repository validators and all PR checks pass.

I found three correctness/integration gaps that need addressing before these rules become authoritative agent guidance. Most importantly, the privacy review skill currently extracts no retention-policy signals and has no targeted cues for either rule. The new article keywords (`retention-policy`, `AddAllowedTable`, `Retention Policy Setup`, install/upgrade codeunits, log/archive tables) do not intersect its current token list, so a reviewer can miss both the absent registration and the “registered but no setup” anti-pattern. Please update `al-privacy-review.md` with relevance tokens and deterministic candidate-selection cues for both articles—for example, accumulating extension-owned log/archive tables or hand-rolled cleanup for the first, and `AddAllowedTable` without a corresponding setup for the second.

The inline findings cover refresh behavior, version scope, and the default-period implementation. With those corrections, the overall contribution should be in good shape.

JesperSchulz (changes requested):
Re-review at `2f3e5afac1fcb4cccf23f09fac1d9e4a79b95e04`: changes are still required. The head is unchanged, and all four prior blockers remain:

- Both articles still use `bc-version: [all]`; retention policies arrived in BC v17 (2020 release wave 2), so use `[17..]`.
- The allowed-table sample still has no `OnRefreshAllowedTables` subscriber, while its upgrade-tag guard prevents refresh from re-registering. Current BCApps installers retain the `ForceUpdate` path.
- The default-policy sample still reimplements period creation instead of public `Retention Policy Setup.FindOrCreateRetentionPeriod(...)`, leaving the documented code-collision failure.
- `al-privacy-review.md` and the fixtures still contain no retention-policy tokens or deterministic routing for either article.

Current BCApps and Learn reconfirm these points. All six required checks pass, and local validation passes (302 articles; 34 fixtures across 17 domains). GitHub reports the PR mergeable but blocked. None of the three unresolved threads is fixed, so I resolved none.

PR #186, opened 2026-09-15
Add Query filter semantics guidance

demiliani:
## Summary

- document that `DataItemTableFilter` remains active when runtime filters target the same field
- document that runtime `SetFilter` and `SetRange` overwrite `ColumnFilter` on the same query column or filter row
- add focused bad/good AL samples and route both checks through the Query review skill
- explicitly cover all Query knowledge articles in the review fixtures

## Validation

- `pwsh ./tools/Test-ReviewFixtures.ps1 -Root .` (70 cases across 18 leaf domains, 8 normalization cases)
- `pwsh ./.github/scripts/Test-KnowledgeIndex.ps1 -Root .` from a non-linked temporary copy (305 articles, 503 samples)
- `git diff --check`
- all four new AL samples compile successfully with AL Compiler 18.0.37.11445 against typed temporary table/enum stubs
- referenced Microsoft Learn page returns HTTP 200 and the expected Business Central content


JesperSchulz (approved):
The two query-filter rules match current Business Central semantics: runtime filters intersect with `DataItemTableFilter`, while `SetFilter`/`SetRange` replace the same column or filter-row `ColumnFilter`. The AL fixtures are focused, all query articles now have positive and clean evaluation cases, and the exact merged tree passes frontmatter, index/retrieval, contract, and 70-case fixture validation. Approving.

The failing custom-layer guard is unrelated to the changes: its GitHub file-list request returned HTTP 502 before path evaluation.

JesperSchulz:
@demiliani, another one for the books! You're on 🔥

demiliani:
I want to fully commit our > 100 devs to BCQuality checks, so trying to move all our best rules/gates here 😅 expect others.

JesperSchulz:
> I want to fully commit our > 100 devs to BCQuality checks, so trying to move all our best rules/gates here 😅 expect others.

I would love to hear how that is going, where things are working well for you, and where you need improvements. Ping me on Teams when you have a minute or 15 one of these weeks please. Sounds like you're running a serious pilot here 🥳

PR #194, opened 2026-09-17
Remove BCQuality logo branding

JesperSchulz:
## Summary

- remove the BCQuality logo block from the root README
- remove the mark and brand-assets link from the documentation index
- delete the logo-only SVGs and brand asset guidance
- restore a text-only README/docs presentation at the user's request

## Validation

- `git diff --check`
- `python .github\scripts\validate_frontmatter.py --root .`
- verified local Markdown and HTML references in `README.md` and `docs/README.md`
- verified no remaining references to `bcq-logo.svg`, `bcq-mark.svg`, `brand-assets.md`, or the deleted BCQuality logo/mark assets


JesperSchulz:
Need to double check if we want to have a logo and branding. We probably don't for a knowledge repo.

sschuh:
But .. to be fair, it looks really sexy 😉

